PR #65, opened 2018-04-19
Provisioners, Worker Types, Workers, Worker

### LINKED ISSUE #15: View provisioners
This is the view synonymous with https://tools.taskcluster.net/provisioners. This is only concerned about the list of provisioners, and not an individual provisioner.

### LINKED ISSUE #16: Worker types
This view is synonymous with https://tools.taskcluster.net/provisioners/aws-provisioner-v1/worker-types. It is concerned with rendering a list of worker types for a particular provisioner. It is also concerned with the actions you can take on a list of worker types. It is not concerned with an individual worker type nor its actions.

### LINKED ISSUE #17: Individual worker type
This view is synonymous with https://tools.taskcluster.net/provisioners/aws-provisioner-v1/worker-types/gecko-1-images. It is concerned with displaying the metadata for a worker type, and its actions. It is not concerned with individual workers, nor their actions.

### LINKED ISSUE #18: Individual worker
This view is synonymous with https://tools.taskcluster.net/provisioners/aws-provisioner-v1/worker-types/gecko-1-images/workers/us-east-1/i-02ca4d161f8d5c3d9. It is concerned with rendering the metadata for a given worker, and its actions. It is not concerned about taking actions on its related tasks.

helfi92:
Resolves #15 
Resolves #16 
Resolves #18 
Resolves #17


helfi92:
## View Provisioners
viewprovisioners

## View Worker Types
viewworkertypes

## View Workers
viewworkers

## View Worker
viewworker
viewworkerinfobutton

src/components/WorkersTable/index.jsx
```diff
@@ -0,0 +1,213 @@
...
+                    </Typography>
+                  }
+                />
+                <ContentCopyIcon />
+              </TableCellListItem>
+            </TableCell>
+            <TableCell>
+              {quarantineUntil
+                ? distanceInWordsStrict(quarantineUntil, Date(), { unit: 'd' })
+                : '-'}
```
eliperelman: Use "n/a".

src/utils/helpers.js
```diff
@@ -0,0 +1,28 @@
+import { compareDesc } from 'date-fns';
```
eliperelman: Rename this file `sort.js` and export sort as default.

src/utils/helpers.js
```diff
@@ -0,0 +1,28 @@
...
+import { or } from 'ramda';
+
+/* eslint-disable import/prefer-default-export */
+
+/**
+ * Return A negative number if the reference element occurs before
```
eliperelman: s/Return A/Return a/

src/utils/prop-types.js
```diff
@@ -66,13 +66,28 @@ export const status = shape({
   runs,
 });
 
-export const task = shape({
-  metadata: shape({
+export const provisionerActions = arrayOf(
```
eliperelman: Can you make a `provisionerAction` of just the shape, and then make this `arrayOf(provisionerAction)`?

src/utils/prop-types.js
```diff
@@ -66,13 +66,28 @@ export const status = shape({
...
+    url: string,
+    method: oneOf(['POST', 'PUT', 'DELETE', 'PATCH']),
     description: string,
-    owner: string,
-    source: string,
-  }),
+  })
+);
+
+export const stability = oneOf[('EXPERIMENTAL', 'STABLE', 'DEPRECATED')];
```
eliperelman: You have the braces inverted here:

```js
export const stability = oneOf(['EXPERIMENTAL', 'STABLE', 'DEPRECATED']);
```

src/utils/prop-types.js
```diff
@@ -89,3 +104,49 @@ export const task = shape({
...
+  recentTasks: arrayOf(
+    shape({
+      taskId: string,
+      runId: number,
+      run,
+    })
+  ),
+  expires: date,
+  quarantineUntil: date,
+  latestTasks: arrayOf(shape(task)),
```
eliperelman: A `task` is already a `shape`:

```js
latestTasks: arrayOf(task),
```

src/views/Provisioners/ViewProvisioners/index.jsx
```diff
@@ -0,0 +1,62 @@
...
+import { withStyles } from 'material-ui/styles';
+import Grid from 'material-ui/Grid';
+import provisionersQuery from './provisioners.graphql';
+import Dashboard from '../../../components/Dashboard';
+import ErrorPanel from '../../../components/ErrorPanel';
+import Spinner from '../../../components/Spinner';
+import ProvisionerDetailsCard from '../../../components/ProvisionerDetailsCard';
+
+@hot(module)
+@graphql(provisionersQuery, {})
```
eliperelman: I think you can remove the empty object.

src/views/Provisioners/ViewProvisioners/index.jsx
```diff
@@ -0,0 +1,62 @@
...
+@withStyles(theme => ({
+  gridItem: {
+    marginBottom: theme.spacing.double,
+  },
+}))
+export default class ViewProvisioners extends PureComponent {
```
eliperelman: Use `Component` here instead of `PureComponent`.

src/views/Provisioners/ViewWorker/index.jsx
```diff
@@ -0,0 +1,59 @@
...
+    },
+  }),
+})
+@withStyles(theme => ({
+  actionButton: {
+    marginRight: theme.spacing.unit,
+    marginBottom: theme.spacing.unit,
+  },
+}))
+export default class ViewWorker extends PureComponent {
```
eliperelman: Use `Component` here instead of `PureComponent`.

src/views/Provisioners/ViewWorker/index.jsx
```diff
@@ -0,0 +1,59 @@
+import { PureComponent, Fragment } from 'react';
+import { hot } from 'react-hot-loader';
```
eliperelman: Import `hot` before `Component`.

src/views/Provisioners/ViewWorker/index.jsx
```diff
@@ -0,0 +1,59 @@
...
+import ErrorPanel from '../../../components/ErrorPanel';
+import WorkerDetailsCard from '../../../components/WorkerDetailsCard';
+import WorkerTable from '../../../components/WorkerTable';
+import workerQuery from './worker.graphql';
+
+@hot(module)
+@graphql(workerQuery, {
+  skip: ({ match: { params } }) => !params.provisionerId,
+  options: ({ match: { params } }) => ({
+    variables: {
```
eliperelman: Would this work by passing `params` as `variables`?

```js
variables: params,
```

src/views/Provisioners/ViewWorkerTypes/index.jsx
```diff
@@ -0,0 +1,152 @@
...
+    variables: {
+      provisionerId,
+      workerTypesConnection: {
+        limit: VIEW_WORKER_TYPES_PAGE_SIZE,
+      },
+      isAwsProvisioner: provisionerId === 'aws-provisioner-v1',
+    },
+  }),
+})
+export default class ViewWorkerTypes extends PureComponent {
```
eliperelman: Use `Component` here instead of `PureComponent`.

src/views/Provisioners/ViewWorkerTypes/index.jsx
```diff
@@ -0,0 +1,152 @@
+import { PureComponent, Fragment } from 'react';
+import { hot } from 'react-hot-loader';
```
eliperelman: Import `hot` before `Component`.

src/views/Provisioners/ViewWorker/worker.graphql
```diff
@@ -0,0 +1,50 @@
...
+    }
+    actions {
+      name
+      title
+      context
+      url
+      description
+    }
+  }
+  workerTypes(provisionerId: $provisionerId, connection: $workerTypesConnection) {
```
eliperelman: Put blank lines in between the function calls just so it's more readable.

src/views/Provisioners/ViewWorkerTypes/index.jsx
```diff
@@ -0,0 +1,152 @@
...
+import workerTypesQuery from './workerTypes.graphql';
+
+@hot(module)
+@withStyles(theme => ({
+  description: {
+    padding: theme.spacing.triple,
+  },
+}))
+@graphql(workerTypesQuery, {
+  skip: ({
```
eliperelman: This would probably be shorter with:

```js
skip: props => !props.match.params.provisionerId,
```

src/views/Provisioners/ViewWorkerTypes/index.jsx
```diff
@@ -0,0 +1,152 @@
...
+      workerTypesConnection: {
+        limit: VIEW_WORKER_TYPES_PAGE_SIZE,
+      },
+      isAwsProvisioner: provisionerId === 'aws-provisioner-v1',
+    },
+  }),
+})
+export default class ViewWorkerTypes extends PureComponent {
+  handleProvisionerChange = ({ target }) => {
+    this.props.history.replace(`/provisioners/${target.value}/worker-types`);
```
eliperelman: Why `replace` instead of `push`?

src/views/Provisioners/ViewWorkerTypes/index.jsx
```diff
@@ -0,0 +1,152 @@
...
+    } = this.props;
+
+    return (
+      <Dashboard
+        title="View Worker Types"
+        user={user}
+        onSignIn={onSignIn}
+        onSignOut={onSignOut}>
+        <Fragment>
+          {!workerTypes && loading && <Spinner loading />}
```
eliperelman: Why the `workerTypes` check here?

src/views/Provisioners/ViewWorkerTypes/index.jsx
```diff
@@ -0,0 +1,152 @@
...
+        error,
+        provisioners,
+        workerTypes,
+        awsProvisionerWorkerTypeSummaries,
+      },
+    } = this.props;
+
+    return (
+      <Dashboard
+        title="View Worker Types"
```
eliperelman: Make this just "Worker Types".

src/views/Provisioners/ViewWorkers/index.jsx
```diff
@@ -0,0 +1,206 @@
...
+    paddingLeft: 0,
+    paddingRight: 0,
+    paddingTop: theme.spacing.double,
+    paddingBottom: theme.spacing.double,
+    '&:last-child': {
+      paddingBottom: theme.spacing.triple,
+    },
+  },
+}))
+export default class ViewWorkers extends PureComponent {
```
eliperelman: Use `Component` here instead of `PureComponent`.

src/views/Provisioners/ViewWorkers/index.jsx
```diff
@@ -0,0 +1,206 @@
+import { hot } from 'react-hot-loader';
+import { PureComponent, Fragment } from 'react';
+import { graphql } from 'react-apollo/index';
```
eliperelman: `/index`?

src/views/Provisioners/ViewWorkers/index.jsx
```diff
@@ -0,0 +1,206 @@
...
+
+  handlePageChange = ({ cursor, previousCursor }) => {
+    const {
+      match: {
+        params: { provisionerId, workerType },
+      },
+      data: { fetchMore },
+    } = this.props;
+
+    return fetchMore({
```
eliperelman: In the future it looks like we can make this more generic in a util.

src/views/Provisioners/ViewWorkers/index.jsx
```diff
@@ -0,0 +1,206 @@
...
+      onSignOut,
+      match: {
+        params: { provisionerId, workerType },
+      },
+      data: { loading, error, workers },
+    } = this.props;
+
+    return (
+      <Dashboard
+        title="View Workers"
```
eliperelman: Make this just "Workers".

src/views/Provisioners/ViewWorkers/index.jsx
```diff
@@ -0,0 +1,206 @@
...
+                      <ListItemText
+                        primary="Actions"
+                        secondary={this.renderActions()}
+                      />
+                    </ListItem>
+                  </List>
+                </CardContent>
+              </Card>
+              <br />
+              {loading ? (
```
eliperelman: Why the loading check here? The connection table already has a loading indicator.

src/views/Provisioners/index.jsx
```diff
@@ -0,0 +1,49 @@
...
+@hot(module)
+export default class Task extends Component {
+  render() {
+    const {
+      match: { path },
+      ...props
+    } = this.props;
+
+    return (
+      <Switch>
```
eliperelman: 😻

src/views/Provisioners/index.jsx
```diff
@@ -0,0 +1,49 @@
...
+);
+const ViewWorker = loadable(() =>
+  import(/* webpackChunkName: 'Provisioners.ViewWorker' */ './ViewWorker')
+);
+const ViewWorkers = loadable(() =>
+  import(/* webpackChunkName: 'Provisioners.ViewWorkers' */ './ViewWorkers')
+);
+
+@hot(module)
+export default class Task extends Component {
```
eliperelman: Rename to `Provisioners` or something.

src/components/WorkerTable/index.jsx
```diff
@@ -0,0 +1,176 @@
...
+    /** A GraphQL worker response. */
+    worker,
+  };
+
+  state = {
+    sortBy: null,
+    sortDirection: null,
+  };
+
+  getTableData() {
```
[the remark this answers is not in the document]
helfi92: Might be interesting. Btw, `memoize` is deprecated in ramda. I'll look into using `memoizeWith`.

src/components/ProvisionerDetailsCard/index.jsx
```diff
@@ -0,0 +1,176 @@
...
+              <LinkIcon />
+            </ListItem>
+            <ListItem
+              button
+              className={classes.listItemButton}
+              onClick={this.handleToggleDescription}>
+              <ListItemText primary="Description" />
+              {showDescription ? <ChevronUpIcon /> : <ChevronDownIcon />}
+            </ListItem>
+            <Collapse in={showDescription} timeout="auto" unmountOnExit>
```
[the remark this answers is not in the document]
helfi92: Agreed. This will, however, require setting a min-height on the card in order to have height consistency with the other cards which should be ok IMO.

src/components/WorkerTable/index.jsx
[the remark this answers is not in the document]
helfi92: When I initially tested it with `memoize`, I noticed that it became slower. I then looked at the implementation and saw that they `toString` the arguments passed to generate the cache key https://github.com/ramda/ramda/blob/master/source/memoize.js#L34. This might explain why it got deprecated :) `memoizeWith` seems to be working fine.

src/components/WorkerTable/index.jsx
```diff
@@ -0,0 +1,176 @@
...
+                    </Typography>
+                  }
+                />
+                <ContentCopyIcon />
+              </TableCellListItem>
+            </TableCell>
+          </TableRow>
+        )}
+        headers={['State', 'Name', 'Task ID', 'Started', 'Resolved']}
+        columnsSize={5}
```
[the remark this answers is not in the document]
helfi92: We still need to be able to support creation of tables without headers. In `ConnectionDataTable` and `DataTable` we can do: `const colSpan = columnsSize || (headers && headers.length) || 1;` and have `columnsSize` optional for when the headers prop is not provided.

src/components/WorkerTypesTable/index.jsx
```diff
@@ -0,0 +1,203 @@
...
+  };
+
+  handleHeaderClick = sortBy => {
+    const toggled = this.state.sortDirection === 'desc' ? 'asc' : 'desc';
+    const sortDirection = this.state.sortBy === sortBy ? toggled : 'desc';
+
+    this.setState({ sortBy, sortDirection });
+  };
+
+  createSortedWorkerTypesConnection = () => {
```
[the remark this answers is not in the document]
helfi92: With `memoizeWith`, it can be:

```js
createSortedWorkerTypesConnection = memoizeWith(
    () =>
      `${toString(this.props.workerTypesConnection.edges)}-${
        this.state.sortBy
      }-${this.state.sortDirection}`,
    () => {
      // ...
    }
);
```

src/utils/prop-types.js
helfi92: Woah, good catch!

src/views/Provisioners/ViewWorker/index.jsx
helfi92: It does!

helfi92:
## Provisioners
screen shot 2018-04-23 at 3 29 37 pm

---

## Worker Types
screen shot 2018-04-23 at 3 29 58 pm

---

## Workers
screen shot 2018-04-23 at 3 30 19 pm

---

## Worker
screen shot 2018-04-23 at 3 30 50 pm

screen shot 2018-04-23 at 3 31 07 pm

screen shot 2018-04-23 at 3 31 23 pm

src/components/SpeedDialActionDialog/index.jsx
```diff
@@ -0,0 +1,129 @@
...
+    } = this.props;
+
+    return (
+      <Fragment>
+        <SpeedDialAction
+          ButtonProps={{ color: 'secondary' }}
+          onClick={this.handleOpen}
+          {...props}
+        />
+        <Dialog open={open} onClose={this.handleClose}>
```
eliperelman: https://material-ui-next.com/demos/dialogs/#responsive-full-screen

src/components/ConnectionDataTable/index.jsx
```diff
@@ -99,50 +129,83 @@ export default class ConnectionDataTable extends Component {
...
-              <TablePagination
-                colSpan={columnsSize}
-                count={count}
-                labelDisplayedRows={Function.prototype}
-                rowsPerPage={pageSize}
-                rowsPerPageOptions={[pageSize]}
-                page={page}
-                onChangePage={this.handlePageChange}
-              />
+      <div>
```
eliperelman: Why multiple divs instead of a Fragment? Why are we rendering 2 tables when loading/why are we changing the existing loading layout?

src/components/ConnectionDataTable/index.jsx
```diff
@@ -99,50 +129,83 @@ export default class ConnectionDataTable extends Component {
...
+          <Table>
+            {headers && (
+              <TableHead>
+                <TableRow>
+                  {headers.map(header => (
+                    <TableCell key={`table-header-${header}`}>
+                      <TableSortLabel
+                        active={header === sortByHeader}
+                        direction={sortDirection || 'desc'}
+                        onClick={() => this.handleHeaderClick(header)}>
```
eliperelman: Can we make this passed by reference, and use an ID to drive the passed value?

```jsx
handleHeaderClick = e => {

};

id=
```

src/components/ButtonDrawer/index.jsx
```diff
@@ -0,0 +1,51 @@
+import { PureComponent, Fragment } from 'react';
+import { node, object } from 'prop-types';
+import Button from 'material-ui/Button';
+import Drawer from 'material-ui/Drawer';
+
+/**
+ * A  button that displays a drawer when clicked.
+ * The drawer opens above all other content until a selection is selected.
+ */
+export default class ButtonDrawer extends PureComponent {
```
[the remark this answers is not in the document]
helfi92: @eliperelman The parent component would be `WorkerTable` and `WorkerTypesTable`. If we put it in the parent component, I think (please correct me if I'm wrong) toggling the `open` state of the drawer will open a drawer for each entry in the table.

src/components/WorkerTable/index.jsx
```diff
@@ -0,0 +1,173 @@
...
+
+@withStyles(theme => ({
+  dateListItem: {
+    marginLeft: -theme.spacing.unit,
+    padding: theme.spacing.unit,
+  },
+  listItemButton: {
+    ...theme.mixins.listItemButton,
+  },
+  taskName: {
```
eliperelman: Can these styles by applied to the cell instead of creating an inner div and applying them?

src/components/WorkerTypeMetadataCard/index.jsx
```diff
@@ -0,0 +1,67 @@
...
+      name: string.isRequired,
+      description: string,
+    }).isRequired,
+  };
+
+  render() {
+    const { classes, metadata } = this.props;
+
+    return (
+      <Card elevation={0}>
```
eliperelman: What purpose does a card serve when its elevation is 0?

src/components/WorkerTypesTable/index.jsx
```diff
@@ -0,0 +1,214 @@
...
+  handleHeaderClick = sortBy => {
+    const toggled = this.state.sortDirection === 'desc' ? 'asc' : 'desc';
+    const sortDirection = this.state.sortBy === sortBy ? toggled : 'desc';
+
+    this.setState({ sortBy, sortDirection });
+  };
+
+  createSortedWorkerTypesConnection = memoizeWith(
+    () =>
+      `${toString(this.props.workerTypesConnection.edges)}-${
```
eliperelman: It appears that a worker type can be uniquely identified using a combination of provisioner ID and worker type name. That would probably give us a better memoization key than using `toString` or `JSON.stringify` (also, pass these in as args; don't access them directly from props and state, which makes memoization pure). They need to be sorted first, otherwise getting the same values in a different order will cause an unnecessary re-calculation.

```js
createSortedWorkerTypesConnection = memoizeWith(
  (
    workerTypesConnection,
    awsProvisionerWorkerTypeSummaries,
    sortBy,
    sortDirection
  ) => {
    const sorted = sortByProvisionerIdAndName(workerTypesConnection.edges);
    const ids = sorted.map(
      ({ provisionerId, workerTypes }) => `${provisionerId}.${workerType}`
    );
    
    return `${ids.join('-')}-${sortBy}-${sortDirection}`;
  },
  (
    workerTypesConnection,
    awsProvisionerWorkerTypeSummaries,
    sortBy,
    sortDirection
  ) => // ...
```

src/components/WorkerTypesTable/index.jsx
```diff
@@ -0,0 +1,214 @@
...
+        this.state.sortBy
+      }-${this.state.sortDirection}`,
+    () => {
+      const { sortBy, sortDirection } = this.state;
+      const {
+        workerTypesConnection,
+        awsProvisionerWorkerTypeSummaries,
+      } = this.props;
+      const sortByProperty = camelCase(sortBy);
+      // Normalize worker types for aws-provisioner-v1
```
eliperelman: Do you think this normalization should happen on the server, or in a util function?

src/components/WorkersTable/index.jsx
```diff
@@ -0,0 +1,230 @@
...
+import sort from '../../utils/sort';
+
+/**
+ * Display relevant information about workers in a table.
+ */
+export default class WorkersTable extends Component {
+  static propTypes = {
+    /** Workers GraphQL PageConnection instance. */
+    workersConnection: shape({
+      edges: array,
```
eliperelman: Can this be typed stronger?

src/components/WorkersTable/index.jsx
```diff
@@ -0,0 +1,230 @@
...
+  };
+
+  state = {
+    sortBy: null,
+    sortDirection: null,
+  };
+
+  createSortedWorkersConnection = memoizeWith(
+    () =>
+      `${toString(this.props.workersConnection.edges)}-${this.state.sortBy}-${
```
eliperelman: See other comment about calculating this.

src/views/Provisioners/ViewWorkerTypes/index.jsx
```diff
@@ -0,0 +1,147 @@
...
+                      </MenuItem>
+                    ))}
+                  </Dropdown>
+                </div>
+                <WorkerTypesTable
+                  workerTypesConnection={workerTypes}
+                  provisionerId={provisionerId}
+                  onPageChange={this.handlePageChange}
+                  awsProvisionerWorkerTypeSummaries={
+                    awsProvisionerWorkerTypeSummaries
```
eliperelman: lol prettier

src/views/Provisioners/index.jsx
```diff
@@ -0,0 +1,49 @@
...
+  render() {
+    const {
+      match: { path },
+      ...props
+    } = this.props;
+
+    return (
+      <Switch>
+        <RouteWithProps
+          path={`${path}/:provisionerId/worker-types/:workerType/workers/:workerGroup?/:workerId?`}
```
eliperelman: I see `workerGroup` and `workerId` are optional, but I couldn't find how their presence affected a view. Could you point me to the right place?

src/components/ConnectionDataTable/index.jsx
helfi92: The outermost `div` can definitely be a Fragment but the second `div` is used to keep the table footer from scrolling with the table. To have `TablePagination` not scroll when the table is small as well as the arrows not positioned at the end of the scroll, it needs to be outside the Table component (see [example](https://material-ui-next.com/demos/tables/#sorting-amp-selecting) in material docs). Next, to display the spinner, a `Table` with a `TableFooter` is created in order to have the spinner positioned at the same spot as the arrows in `TablePagination`.

To stop rendering 2 tables, the second Table can be replaced with:

```js
<div className={classes.spinner}>
  <Spinner size={24} />
</div>
```

```js
spinner: {
  height: 56,
  minHeight: 56,
  paddingRight: 2,
  display: 'flex',
  alignItems: 'center',
  flexDirection: 'row-reverse',
},
```
I think ^ makes more sense to use.

src/components/ConnectionDataTable/index.jsx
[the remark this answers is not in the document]
helfi92: We could, but I think we should (when we can) try to use IDs for anchors only. We could rely on `e.target.textContent` instead, if that's ok with you.

src/components/DataTable/index.jsx
```diff
@@ -0,0 +1,114 @@
...
+  TableBody,
+  TableCell,
+  TableHead,
+  TableSortLabel,
+  TableRow,
+} from 'material-ui/Table';
+
+@withStyles({
+  tableWrapper: {
+    overflowX: 'auto',
```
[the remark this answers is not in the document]
helfi92: Weird, it seems to be working now with our current `1.0.0-beta.41`. I'll just remove it.

src/components/ConnectionDataTable/index.jsx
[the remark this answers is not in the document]
eliperelman: > try to use IDs for anchors only

Why?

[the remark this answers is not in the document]
helfi92: Sorry, by anchors I mean things like toolbar, header, footer, etc. I would typically use `id`s as fragment identifier to jump to anchors. There's also the specificity war that can happen with CSS. I don't think it's the end of the world if we use an `id` here but using something like `textContent` will avoid using an `id` and is altogether less code.

src/components/QuarantineSpeedDialAction/index.jsx
```diff
@@ -0,0 +1,76 @@
...
+
+    this.state = {
+      quarantineUntil: props.quarantineUntil
+        ? new Date(props.quarantineUntil)
+        : addYears(new Date(), 1000),
+    };
+  }
+
+  handleQuarantineClick = () => {
+    this.props.onQuarantineClick(format(this.state.quarantineUntil));
```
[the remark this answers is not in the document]
helfi92: It passes a string.

`this.state.quarantineUntil` is a Date object
`format(this.state.quarantineUntil)` is a string

src/components/WorkerTable/index.jsx
helfi92: @eliperelman There's two elements inside `TableCellListItem`. It would be hard to target the text node without wrapping it in a `div`. I'm open to suggestions.

src/components/WorkerTypesTable/index.jsx
helfi92: I think a util function makes more sense here. We have 14 provisioners and this normalization is only for `aws-provisioner`.

src/views/Provisioners/index.jsx
helfi92: This needs to be removed. We don't have dropdowns in the worker view anymore since most users come to this page via the workers view. I will revise.

src/components/QuarantineSpeedDialAction/index.jsx
```diff
@@ -0,0 +1,76 @@
...
+        body={
+          <Fragment>
+            <Fragment>
+              Quarantining a worker allows the machine to remain alive but not
+              accept jobs. Note that a quarantine can be lifted by setting
+              Quarantine Until to the present time or somewhere in the past.
+            </Fragment>
+            <br />
+            <br />
+            <TextField
```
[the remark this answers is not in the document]
helfi92: @eliperelman I tried using `material-ui-pickers` but having two modals open at the same time is not so pretty. Also, I want to point out that I've been having trouble making the `onChange` handler work on the `material-ui-pickers` component. I suspect it has something to do with the `z-index` since 2 modals are open. tbh, I think, for now, leaving the current basic implementation should be okay. We can revisit later.

screen shot 2018-04-25 at 3 50 05 pm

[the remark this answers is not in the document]
helfi92: The current implementation is pretty ugly as well...ugh

helfi92:
## Provisioners

screen shot 2018-04-26 at 3 09 01 pm

---

## Worker Types

screen shot 2018-04-26 at 3 09 31 pm

---

## Workers

screen shot 2018-04-26 at 3 09 51 pm

---

## Worker

screen shot 2018-04-26 at 3 10 22 pm

src/components/DateDistance/index.jsx
```diff
@@ -26,8 +26,8 @@ export default class DateDistance extends Component {
 
   render() {
     const { from, offset } = this.props;
-    const fromNow = distanceInWordsToNow(from, { addSuffix: true });
-    const offsetNow = offset && distanceInWords(offset, from);
+    const fromNow = formatDistanceStrict(from, new Date(), { addSuffix: true });
+    const offsetNow = offset && formatDistance(from, offset);
```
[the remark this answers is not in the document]
helfi92: Yeah, they were renamed and the arguments order was switched.  The change is well explained in the [`CHANGELOG`](https://github.com/date-fns/date-fns/blob/master/CHANGELOG.md).

src/components/WorkerTypesTable/index.jsx
```diff
@@ -0,0 +1,266 @@
...
+          onHeaderClick={this.handleHeaderClick}
+          onPageChange={onPageChange}
+          headers={headers}
+          renderRow={({ node: workerType }) => (
+            <TableRow key={workerType.workerType}>
+              <TableCell>
+                <Button
+                  className={classes.infoButton}
+                  size="small"
+                  onClick={() => this.handleDrawerOpen(workerType)}>
```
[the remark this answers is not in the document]
helfi92: @eliperelman `workerType` is an object here, not a string. How would you gain reference to the object without having to loop over the sorted worker types connection?

src/components/WorkersTable/index.jsx
```diff
@@ -0,0 +1,241 @@
...
+                      <DateDistance from={firstClaim} />
+                    </Typography>
+                  }
+                />
+                <ContentCopyIcon size={iconSize} />
+              </TableCellListItem>
+            </TableCell>
+            <TableCell>
+              {quarantineUntil
+                ? formatDistanceStrict(Date(), quarantineUntil, { unit: 'd' })
```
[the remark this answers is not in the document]
helfi92: The parameter can be of type `Date | String | Number` https://date-fns.org/v2.0.0-alpha.7/docs/formatDistanceStrict#arguments. I'll be sure to be consistent everywhere.

src/components/WorkerTypesTable/index.jsx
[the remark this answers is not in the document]
eliperelman: You *would* need to loop over the sorted connection to look up the worker type by the ID.

src/components/DateDistance/index.jsx
```diff
@@ -26,7 +26,7 @@ export default class DateDistance extends Component {
 
   render() {
     const { from, offset } = this.props;
-    const fromNow = formatDistanceStrict(from, new Date(), { addSuffix: true });
+    const fromNow = formatDistanceStrict(from, Date(), { addSuffix: true });
```
eliperelman: Change `Date` to `new Date()`.

src/views/Provisioners/ViewWorkers/index.jsx
```diff
@@ -143,25 +149,24 @@ export default class ViewWorkers extends Component {
...
+                  <SpeedDial>
+                    {workerType.actions.map(action => (
+                      <SpeedDialAction
+                        key={action.title}
+                        ButtonProps={{ color: 'secondary' }}
+                        icon={<HammerIcon />}
+                        tooltipTitle={
+                          <div>
+                            <div>{action.title}</div>
+                            <div>{action.description}</div>
```
eliperelman: Remove description. The description will be in the dialog that gets opened, and putting it here will be potentially too large and redundant to show every time.

src/views/Provisioners/ViewWorkers/index.jsx
```diff
@@ -143,25 +149,24 @@ export default class ViewWorkers extends Component {
...
+                          </div>
+                        }
+                        onClick={() => this.handleActionClick(action)}
```
eliperelman: We can leave this for now if it keeps it simple. Maybe add a TODO to come back later.

src/components/WorkerTypesTable/index.jsx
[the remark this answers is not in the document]
helfi92: Wouldn't this make it O(n)? The current implementation is O(1).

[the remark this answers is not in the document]
eliperelman: It also creates n event handlers instead of 1.

PR #70, opened 2018-05-14
Add clientId textfield when creating a new client

helfi92:
The client ID field should be editable when creating a new client but not when editing an existing client. This commit implements this logic.

### DIFF
--- a/src/components/ClientForm/index.jsx
+++ b/src/components/ClientForm/index.jsx
@@ -143,7 +143,7 @@ export default class ClientForm extends Component {
               />
             </FormGroup>
           </ListItem>
-          {isNewClient ? (
+          {isNewClient && (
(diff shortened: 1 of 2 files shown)


src/components/ClientForm/index.jsx
```diff
@@ -143,6 +143,17 @@ export default class ClientForm extends Component {
               />
             </FormGroup>
           </ListItem>
+          {isNewClient ? (
```
eliperelman: `{isNewClient && (`

helfi92: Wouldn't this actually display "False" to the view when it's not a new client?

helfi92: OK, maybe not :)

Issue #45, opened 2018-04-10
Create/edit hook


eliperelman:
This view is synonymous with https://tools.taskcluster.net/hooks/garbage/12345. We should show a pre-filled form when loading an existing hook, and blank when creating a new hook. No need for a view-only mode, just the form view.

Issue #44, opened 2018-04-10
Hooks list


eliperelman:
This view is synonymous with https://tools.taskcluster.net/hooks. We are only concerned with the list of hooks here, not an individual hook.

PR #71, opened 2018-05-23
Hooks list & create/edit/view a hook

helfi92:
Resolves https://github.com/taskcluster/taskcluster-web/issues/44.
Resolves https://github.com/taskcluster/taskcluster-web/issues/45.


helfi92:
### Hooks
Searching will filter the tree view as well.

screen shot 2018-05-23 at 4 12 46 pm

### Create Hook

screen shot 2018-05-23 at 4 12 13 pm

### View/Edit Hook

screen shot 2018-05-23 at 4 13 21 pm

src/components/Tree/index.jsx
```diff
@@ -0,0 +1,193 @@
+import { Component } from 'react';
```
helfi92: I'm thinking of extracting this component into its own library and let other material users be able to use it. What do you think?

eliperelman: Absolutely.

src/components/Tree/index.jsx
```diff
@@ -0,0 +1,193 @@
...
+import { arrayOf, shape, string, func, oneOfType, object } from 'prop-types';
+import { memoizeWith } from 'ramda';
+import { withStyles } from 'material-ui/styles';
+import { ListItem } from 'material-ui/List';
+import ExpansionPanel, {
+  ExpansionPanelSummary,
+  ExpansionPanelDetails,
+} from 'material-ui/ExpansionPanel';
+import ChevronDownIcon from 'mdi-react/ChevronDownIcon';
+import theme from '../../theme';
```
eliperelman: If you need the theme in the component, pass `withTheme: true` to the options and it will become a prop.

```js
@withStyles(
  theme => ({
    // styles...
  }),
  {
    // options
    withTheme: true,
  }
)
export default class Tree extends Component {
  method() {
    this.props.theme
  }
}
```

src/components/Tree/index.jsx
```diff
@@ -0,0 +1,193 @@
...
+        <ExpansionPanelSummary
+          className={classes.panelSummary}
+          expandIcon={<ChevronDownIcon />}
+          {...expansionPanelSummaryProps}>
+          <div className={classes.text}>{node.value}</div>
+        </ExpansionPanelSummary>
+        <ExpansionPanelDetails
+          className={classes.panelDetails}
+          {...expansionPanelDetailsProps}>
+          {(treeDepth += 1) &&
```
eliperelman: This mutates `treeDepth`. What does this do?

src/components/Tree/index.jsx
```diff
@@ -0,0 +1,193 @@
...
+            node.nodes.map(l => this.renderNode(l, node, treeDepth))}
+        </ExpansionPanelDetails>
+      </ExpansionPanel>
+    );
+  };
+
+  isLeaf(node) {
+    if (typeof node === 'string' || !node.nodes || !node.nodes.length) {
```
eliperelman:
```js
return typeof node === 'string' || !node.nodes || !node.nodes.length;
```

src/components/Tree/index.jsx
```diff
@@ -0,0 +1,193 @@
...
+      return true;
+    }
+
+    return false;
+  }
+
+  getNodeValue(node) {
+    if (typeof node === 'string') {
```
eliperelman: This can be a ternary:

```js
return typeof node === 'string'
  ? node
  : node.value;
```

src/components/Tree/index.jsx
```diff
@@ -0,0 +1,193 @@
...
+      return node;
+    }
+
+    return node.value;
+  }
+
+  filter(tree) {
+    const { searchTerm } = this.props;
+    const newTree = tree.filter(node => {
```
eliperelman: `newTree` isn't used, so just return it:

```js
return tree.filter(node => {
```

src/views/Hooks/ViewHook/hook.graphql
```diff
@@ -0,0 +1,49 @@
...
+        name
+        description
+        owner
+        source
+      }
+      tags
+      extra
+    }
+  }
+}
```
eliperelman: Newline.

src/views/Hooks/ViewHooks/hooks.graphql
```diff
@@ -0,0 +1,8 @@
+query Hooks {
+  hookGroups {
+    hookGroupId
+    hooks {
+      hookId
+    }
+  }
+}
```
eliperelman: Newline.

src/views/Hooks/index.jsx
```diff
@@ -0,0 +1,39 @@
+import { hot } from 'react-hot-loader';
+import { Component } from 'react';
+import { Switch } from 'react-router-dom';
+import RouteWithProps from '../../components/RouteWithProps';
+import loadable from '../../utils/loadable';
+
+const ViewHooks = loadable(() =>
+  import(/* webpackChunkName: 'Clients.ViewHooks' */ './ViewHooks')
```
eliperelman: Can we make this one `Hooks.ListHooks` for a `ListHooks` so it's a little more distinct from the single hook page?

src/views/Hooks/index.jsx
```diff
@@ -0,0 +1,39 @@
...
+);
+const ViewHook = loadable(() =>
+  import(/* webpackChunkName: 'Clients.ViewHook' */ './ViewHook')
```
eliperelman: `Hooks.ViewHook`?

src/components/CodeEditor/styles.css
```diff
@@ -0,0 +1,4 @@
+:global .CodeMirror {
+    font-size: 13px;
```
eliperelman: You can put this in the code editor component, I'll comment there.

src/components/HookForm/index.jsx
```diff
@@ -0,0 +1,464 @@
...
+                  When the hook fires, this template is rendered with{' '}
+                  <a
+                    href="https://taskcluster.github.io/json-e/"
+                    target="_blank"
+                    rel="noopener noreferrer">
+                    JSON-e
+                  </a>{' '}
+                  to create the the task definition. See{' '}
+                  <a
+                    href="https://docs.taskcluster.net/reference/core/taskcluster-hooks/docs/firing-hooks"
```
eliperelman: Can we bring in `taskcluster-lib-urls` for this?

src/components/HookForm/index.jsx
```diff
@@ -0,0 +1,464 @@
...
+                options={{ mode: 'json' }}
+                value={JSON.stringify(task, null, 2)}
+                onChange={this.handleTaskChange}
+              />
+            </ListItem>
+          </List>
+          <List subheader={<ListSubheader>Trigger Schema *</ListSubheader>}>
+            <ListItem className={classes.editorListItem}>
+              <Typography variant="caption">
+                <span>
```
eliperelman: Do we need this span?

src/components/CodeEditor/index.jsx
```diff
@@ -0,0 +1,67 @@
...
+import 'codemirror/mode/yaml/yaml';
+import 'codemirror/addon/display/placeholder';
+import 'codemirror/lib/codemirror.css';
+import 'codemirror/addon/lint/lint.css';
+import 'codemirror/theme/material.css';
+import './yaml-lint';
+import './styles.css';
+
+@withStyles({
+  codeMirror: {
```
eliperelman: Add global styles with:

```js
@withStyles({
  '@global .CodeMirror': {
    font-size: 13,
    height: '100%',
  },
})
```

src/components/CodeEditor/yaml-lint.js
```diff
@@ -0,0 +1,26 @@
+import CodeMirror from 'codemirror';
```
eliperelman: Can you switch to `react-codemirror2`?

Something similar:

```jsx
import { Component } from 'react';
import { bool, func, shape, string } from 'prop-types';
import { withStyles } from '@material-ui/core/styles';
import { Controlled } from 'react-codemirror2';
import 'codemirror/mode/xml/xml';
import 'codemirror/mode/javascript/javascript';
import 'codemirror/lib/codemirror.css';
import 'codemirror/theme/material.css';

@withStyles({
  root: {
    fontSize: 16,
  },
})
export default class CodeEditor extends Component {
  static propTypes = {
    value: string.isRequired,
    onChange: func,
    options: shape({
      mode: string,
      theme: string,
      lineNumbers: bool,
    }),
  };

  static defaultProps = {
    onChange: null,
    options: {
      theme: 'material',
      lineNumbers: true,
    },
  };

  handleTextUpdate = (editor, data, value) => {
    if (this.props.onChange) {
      this.props.onChange(value);
    }
  };

  render() {
    const { value, options, classes } = this.props;

    return (
      <Controlled
        className={classes.root}
        value={value}
        options={options}
        onBeforeChange={this.handleTextUpdate}
      />
    );
  }
}

```

package.json
```diff
@@ -17,6 +17,7 @@
   "dependencies": {
     "@material-ui/icons": "^1.0.0-beta.43",
     "@material-ui/lab": "^1.0.0-alpha.0",
+    "@skidding/react-codemirror": "^1.0.2",
```
eliperelman: Switch to `react-codemirror2`. I commented with a usage of this I have done before.

src/components/CodeEditor/index.jsx
helfi92: That doesn't seem to work even after changing `font-size` to `fontSize`.

src/components/CodeEditor/yaml-lint.js
helfi92: Absolutely!

src/components/HookForm/index.jsx
helfi92: Sure, but I think the generated link won't work right away.

src/components/Tree/index.jsx
helfi92: No idea what I was doing here. Fixing this now.

helfi92:
Added the requested changes. 2 things I wanted to bring up:

* Using '@global .CodeMirror' did not work.
* The url generated using `taskcluster-lib-urls` will not exactly match with the current setup of docs.  I think that's ok seeing we won't deploy this site this quarter.

eliperelman:
> The url generated using `taskcluster-lib-urls` will not exactly match with the current setup of docs.  I think that's ok seeing we won't deploy this site this quarter.

Can you elaborate on what the problem is here?

eliperelman:
Oh, nevermind, I see. This is probably because we aren't passing the right `TASKCLUSTER_ROOT` or whatever yet.

src/components/CodeEditor/index.jsx
eliperelman: Hmm weird, this should have worked.

.neutrinorc.js
```diff
@@ -65,6 +65,7 @@ module.exports = {
       'AUTH0_RESPONSE_TYPE',
       'AUTH0_SCOPE',
       'BASE_URL',
+      'DOMAIN',
```
eliperelman: We already have `BASE_URL` which is supposed to drive this, but it has been renamed `TASKCLUSTER_ROOT_URL`.

src/components/CodeEditor/index.jsx
```diff
@@ -0,0 +1,62 @@
...
+
+  render() {
+    const { classes, value, onChange: _, ...options } = this.props;
+    const opts = {
+      mode: 'application/json',
+      theme: 'material',
+      indentWithTabs: false,
+      gutters: ['CodeMirror-lint-markers'],
+      lineNumbers: true,
+      ...options,
```
eliperelman: I'm guessing linting is dying on this line because there is a trailing comma after a spread.

src/components/CodeEditor/yaml-lint.js
```diff
@@ -0,0 +1,26 @@
...
+import { safeLoad } from 'js-yaml';
+import 'codemirror/addon/lint/lint';
+
+/*
+  Override 'codemirror/addon/lint/yaml-lint' registerHelper
+  so that it doesn't use window.jsyaml
+  https://github.com/codemirror/CodeMirror/blob/master/addon/lint/yaml-lint.js
+ */
+CodeMirror.registerHelper('lint', 'yaml', text => {
```
eliperelman: This could be rewritten as:

```js
CodeMirror.registerHelper('lint', 'yaml', text => {
  try {
    safeLoad(text);
    return [];
  } catch (e) {
    const loc = e.mark;
 
    return [{
      from: CodeMirror.Pos(loc.line, loc.column),
      to: CodeMirror.Pos(loc.line, loc.column),
      message: e.message,
    }];
  }
});
```

src/components/Tree/index.jsx
```diff
@@ -0,0 +1,186 @@
...
+  ExpansionPanelDetails,
+} from 'material-ui/ExpansionPanel';
+import ChevronDownIcon from 'mdi-react/ChevronDownIcon';
+
+/** Prop-type for a recursive data structure */
+const tree = {
+  value: string.isRequired,
+};
+
+tree.nodes = arrayOf(oneOfType([shape(tree), string]));
```
eliperelman: nit: use `Object.assign`. 🤷‍♂️

src/components/CodeEditor/styles.css
eliperelman: Try it again with something like this:

https://github.com/taskcluster/taskcluster-web/blob/master/src/App/index.jsx#L23

```js
'@global': {
  '.CodeMirror': {
    fontSize: 13,
    height: '100%',
  },
},
```

src/components/CodeEditor/index.jsx
helfi92: The linter complains if I remove it. Curious, why is it wrong to have a trailing comma after a spread?

eliperelman:
Something is wrong with the Netlify build. :(

helfi92:
Oh no! Would you say the issue coming from this PR or the netlify configs?

eliperelman:
It could be the PR, lemme check the deps changes.

eliperelman:
Can you push a commit which just removes taskcluster-lib-urls and see what happens?

helfi92:
That doesn't seem to have been the issue. Hmm...

helfi92:
Seems to be a netlify issue. Hopefully neutrino v9 will fix this. Let me know if there's anything else to change.

eliperelman:
Ship it.

Issue #32, opened 2018-04-10
AWS provisioner health


eliperelman:
This view is synonymous with https://tools.taskcluster.net/aws-provisioner/aws-health.

PR #73, opened 2018-06-01
AWS provisioner health view

helfi92:
Resolves https://github.com/taskcluster/taskcluster-web/issues/32.

### DIFF
--- a/src/components/AwsProvisionerHealthTable/index.jsx
+++ b/src/components/AwsProvisionerHealthTable/index.jsx
@@ -3,9 +3,7 @@ import {
   memoizeWith,
   pipe,
   map,
-  ifElse,
   isEmpty,
-  identity,
   defaultTo,
   sort as rSort,
 } from 'ramda';
@@ -33,6 +31,8 @@ const sorted = pipe(
   ),
   map(({ region, az, instanceType }) => `${region}-${az}-${instanceType}`)
 );
+const iconSize = 18;
+const or0 = defaultTo(0);
 
 @withStyles(theme => ({
   drawerContainer: {
@@ -75,7 +75,6 @@ export default class AwsProvisionerHealthTable extends Component {
       const { requestHealth, terminationHealth, running } = healthData;
       const identifier = item =>
         `${item.az}-${item.region}-${item.instanceType}`;
-      const or0 = defaultTo(0);
 
       requestHealth.forEach(item => {
         healthSummary[identifier(item)] = {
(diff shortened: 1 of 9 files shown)


helfi92:
This PR relies on https://github.com/taskcluster/taskcluster-web-server/pull/9 being merged first.

src/components/AwsProvisionerHealthTable/index.jsx
```diff
@@ -0,0 +1,296 @@
...
+        identity,
+        rSort((a, b) => {
+          const firstElement =
+            sortDirection === 'desc' ? b[sortByProperty] : a[sortByProperty];
+          const secondElement =
+            sortDirection === 'desc' ? a[sortByProperty] : b[sortByProperty];
+
+          return sort(firstElement, secondElement);
+        })
+      )(healthValues);
```
eliperelman: If we aren't using point-free syntax, there is little value to an `ifElse` call like this and can be more easily read imperatively.

src/components/AwsProvisionerHealthTable/index.jsx
```diff
@@ -0,0 +1,296 @@
...
+  render() {
+    const { classes, healthData } = this.props;
+    const { sortBy, sortDirection, drawerOpen, drawerItem } = this.state;
+    const sortedHealth = this.createSortedHealth(
+      healthData,
+      sortBy,
+      sortDirection
+    );
+    const iconSize = 18;
+    const or0 = defaultTo('0');
```
eliperelman: This isn't the only place `or0` is used so hoist it to the top.

src/components/AwsProvisionerHealthTable/index.jsx
```diff
@@ -0,0 +1,296 @@
...
+
+  render() {
+    const { classes, healthData } = this.props;
+    const { sortBy, sortDirection, drawerOpen, drawerItem } = this.state;
+    const sortedHealth = this.createSortedHealth(
+      healthData,
+      sortBy,
+      sortDirection
+    );
+    const iconSize = 18;
```
eliperelman: This can be a constant at the top since it doesn't need to change with each render.

src/views/AwsProvisioner/index.jsx
```diff
@@ -0,0 +1,29 @@
+import { hot } from 'react-hot-loader';
+import { Component } from 'react';
+import { Switch } from 'react-router-dom';
+import RouteWithProps from '../../components/RouteWithProps';
+import loadable from '../../utils/loadable';
+
+const ViewAwsHealth = loadable(() =>
+  import(/* webpackChunkName: 'Clients.ViewClients' */ './ViewAwsHealth')
```
eliperelman: s/Clients/AwsProvisioner/.

src/views/AwsProvisioner/index.jsx
```diff
@@ -0,0 +1,29 @@
...
+);
+
+@hot(module)
+export default class Clients extends Component {
```
eliperelman: Not `Clients`.

src/views/AwsProvisioner/index.jsx
```diff
@@ -0,0 +1,29 @@
...
+    const {
+      match: { path },
+      ...props
+    } = this.props;
+
+    return (
+      <Switch>
+        <RouteWithProps
+          path={`${path}/aws-health`}
+          {...props}
```
eliperelman: What props need to be spread from view to view, and only a single view?

helfi92: I'm not sure I understand what you mean, but if you're asking whether `/aws-health` will need any props from this file, then probably not. I will remove it.

src/components/AwsProvisionerHealthTable/index.jsx
helfi92: Agreed.

src/views/AwsProvisioner/index.jsx
helfi92: Nevermind, `onSignIn`, `onSignOut`, and `user`are needed in the component.

eliperelman: Ah I see. I'll have that fixed in my auth PR.

PR #74, opened 2018-06-18
Add aws provisioner worker types list

### LINKED ISSUE #27: AWS provisioner list
This page is synonymous with https://tools.taskcluster.net/aws-provisioner. I'm still not a fan of giving the AWS provisioner a special page over other provisioners, but it is what it is. This view is only concerned with the data and actions of the provisioner itself, and is not concerned with individual worker types, their actions, creating worker types, recent errors or health. Those may be linked, but they will not render in this view.

helfi92:
Resolves #27.


helfi92:
screen shot 2018-06-18 at 3 40 15 pm

src/theme.js
```diff
@@ -99,9 +108,27 @@ const theme = createMuiTheme({
...
-      backgroundColor: SUCCESS.MAIN,
+      backgroundColor: success.main,
+      '&:hover': {
+        backgroundColor: success.dark,
+      },
+      '& svg': {
+        backgroundColor: 'transparent',
+      },
+    },
+    yellowIcon: {
```
eliperelman: I'd like to see these named by intent rather than color. So `successIcon`, `warningIcon`, and `errorIcon`, so they align with the palettes.

src/views/AwsProvisioner/ViewWorkerTypes/workerTypes.graphql
```diff
@@ -0,0 +1,8 @@
+query workerTypes {
+  awsProvisionerWorkerTypeSummaries {
+    workerType
+    pendingCapacity
+    runningCapacity
+    pendingTasks
+  }
+}
```
eliperelman: New line.

PR #80, opened 2018-06-27
Add Quickstart view

helfi92:
Resolves #80.

### DIFF
--- a/src/App/routes.jsx
+++ b/src/App/routes.jsx
@@ -89,7 +89,7 @@ export default [
   },
   {
     component: loadable(() =>
-      import(/* webpackChunkName: 'CachePurges' */ '../views/Quickstart')
+      import(/* webpackChunkName: 'Quickstart' */ '../views/Quickstart')
     ),
     path: '/quickstart',
   },
--- a/src/App/routes.jsx
+++ b/src/App/routes.jsx
@@ -89,6 +89,12 @@ export default [
   },
   {
     component: loadable(() =>
+      import(/* webpackChunkName: 'CachePurges' */ '../views/Quickstart')
+    ),
+    path: '/quickstart',
+  },
+  {
+    component: loadable(() =>
       import(/* webpackChunkName: 'NotFound' */ '../views/NotFound')
     ),
   },
--- a/src/views/Quickstart/github.graphql
+++ b/src/views/Quickstart/github.graphql
@@ -1,5 +1,5 @@
-query Github($owner: String!, $repo: String!) {
-  githubRepository(owner:$owner, repo: $repo) {
+query GitHub($owner: String!, $repo: String!) {
+  githubRepository(owner: $owner, repo: $repo) {
     installed
   }
 }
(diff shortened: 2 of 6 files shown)


helfi92:
screen shot 2018-06-27 at 4 19 15 pm

eliperelman:
In the current UI there is a radio since the trigger option is a binary choice. Using checkboxes is a visual indicator that more than one option can be chosen. I'd suggest either using a switch or a dropdown:

switch

src/App/routes.jsx
```diff
@@ -87,6 +87,12 @@ export default [
     ),
     path: '/purge-caches',
   },
+  {
+    component: loadable(() =>
+      import(/* webpackChunkName: 'CachePurges' */ '../views/Quickstart')
```
eliperelman: s/CachePurges/QuickStart/

src/views/Quickstart/github.graphql
```diff
@@ -0,0 +1,5 @@
+query Github($owner: String!, $repo: String!) {
```
eliperelman: Name this query "GitHub".

src/views/Quickstart/github.graphql
```diff
@@ -0,0 +1,5 @@
...
+  githubRepository(owner:$owner, repo: $repo) {
```
eliperelman: Space after `:`.

src/views/Quickstart/index.jsx
```diff
@@ -0,0 +1,469 @@
...
+
+  handleCommandsChange = ({ target: { value } }) => {
+    this.setState({
+      commandSelection: value,
+      commands: value === 'standard' ? cmdDirectory(this.state.image) : [],
+      editorValue: null,
+    });
+  };
+
+  getInstalledState = debounce(async (owner, repo) => {
```
eliperelman: I think this is the ideal situation where a `<Query>` component makes sense. We should be able to use the state to drive the variables passed to the Query and use a render prop to drive the rendering based on that. Make sense?

helfi92: When the `Query` component is rendered it will automatically fire off the query which is not my intention to do. I went with the manual approach so that I could make sure the request uses a debouncer for the query call. For delaying queries, Apollo [suggests](https://www.apollographql.com/docs/react/essentials/queries.html#manual-query) using `client.query()`. How would you use the `Query` component while making sure the query is debounced?

src/App/routes.jsx
helfi92: Ugh, I keep doing the same silly mistake.

helfi92:
"Collaborators" is not the opposite of "Public" so I don't think a switch would sit nicely here. I will use a dropdown instead. I agree that checkboxes for this use case is not the best UX.

src/views/Quickstart/index.jsx
eliperelman: Wouldn't you just render the component or not based on the value of the props, i.e. if the props weren't there, don't render the Query component? If not, feel free to move forward with this approach, just let me know which way so I can review or not. :)

helfi92: Ah, `getInstalledState` is actually triggered everytime a user types something inside the repository or organization text field. It is being called by `handleOrgRepoChange`. I will move forward with this approach :)

helfi92:
😍

PR #84, opened 2018-07-20
Add Toggeable light theme

### LINKED ISSUE #50: Toggleable light theme
In addition to the dark theme, we can put a preference in the user's profile (which maybe we could rename to settings, not sure), to toggle between a light and dark theme.

helfi92:
Resolves #50.


src/App/withThemeToggler.jsx
```diff
@@ -0,0 +1,34 @@
+import { Component } from 'react';
+import { MuiThemeProvider } from '@material-ui/core/styles';
+import theme from '../theme';
+
+/** A HOC that provides the ability toggle themes  */
+const withThemeToggler = ComposedComponent =>
+  class ThemeToggler extends Component {
+    state = {
+      theme: theme.lightTheme,
```
eliperelman: This sets the default theme to light correct? We want dark to be the default.

src/App/withThemeToggler.jsx
```diff
@@ -0,0 +1,34 @@
+import { Component } from 'react';
+import { MuiThemeProvider } from '@material-ui/core/styles';
+import theme from '../theme';
+
+/** A HOC that provides the ability toggle themes  */
+const withThemeToggler = ComposedComponent =>
```
eliperelman: You should be able to default export this:

```js
export default ComposedComponent =>
```

src/components/Dashboard/index.jsx
```diff
@@ -39,6 +42,7 @@ import { user } from '../../utils/prop-types';
     appBarTitle: {
       fontFamily: 'Roboto300',
       flex: 1,
+      color: 'rgba(255, 255, 255, 0.9)',
```
eliperelman: Since this is close to white, should this switch based on theme?

src/components/Dashboard/index.jsx
```diff
@@ -60,10 +64,12 @@ import { user } from '../../utils/prop-types';
         position: 'fixed',
       },
       borderRight: 0,
+      backgroundColor:
+        theme.palette.type === 'dark' ? theme.palette.primary.main : '#fafafa',
```
eliperelman: Wouldn't the light theme set its own `palette.primary.main`?

src/views/NotFound/Emoticon.jsx
```diff
@@ -0,0 +1,17 @@
+const Emoticon = ({ fill, className }) => {
```
eliperelman: I know it seems lame, but can you make this a `Component`/`PureComponent`?

eliperelman:
Can you post some screenshits in the interim?

helfi92:
screen shot 2018-07-25 at 3 16 15 pm

---

screen shot 2018-07-25 at 3 17 01 pm

---

screen shot 2018-07-25 at 3 17 28 pm

---

screen shot 2018-07-25 at 3 18 58 pm

src/components/Dashboard/index.jsx
helfi92: The app bar title sits behind the blue background which doesn't change on theme toggle, hence the hardcoded value.

eliperelman:
Such hotness! Can you move the theme toggle to the left side of the search?

src/components/Dashboard/index.jsx
helfi92: You're absolutely right. I will send a fix :)

helfi92:
Some views don't have a search bar. Because of that, having the theme toggle icon move from the left of the search component to the far right when there's no search component is not super consistent IMHO. Let me know if you disagree.

eliperelman:
OK, I'm fine with keeping it there for now, with the intent that some other icons will also need to be there in the future, such as for user settings, etc. I just don't want to give the theme toggler too much prominence.

eliperelman:
Can you explain why the theme toggler exists in a decorator instead of just being a part of the state of `App`?

helfi92:
Definitely, I had forgotten about this feature.

helfi92:
Done.

helfi92:
@eliperelman Do you have anything else to recommend for this pull-request?

Issue #43, opened 2018-04-10
Task Index


eliperelman:
This view is synonymous with https://tools.taskcluster.net/index. I hate the UX on this page, so we need some thought here on how to make it better. Maybe a stepper wizard or something. 🤷‍♂️

PR #86, opened 2018-08-02
Add task index view

helfi92:
Resolves https://github.com/taskcluster/taskcluster-web/issues/43.

### DIFF
--- a/src/utils/buildArtifactUrl.js
+++ b/src/utils/buildArtifactUrl.js
@@ -1,0 +1,49 @@
+import { isNil } from 'ramda';
+import getIconFromMime from './getIconFromMime';
+import urls from './urls';
+
+export default ({ name, contentType, namespace, runId, taskId }) => {
+  if (/^public\//.test(name)) {
+    const icon = getIconFromMime(contentType);
+
+    // If we have a namespace, use a URL with that namespace to make it
+    // easier for users to copy/paste index URLs
+    if (namespace) {
+      return {
+        icon,
+        name,
+        url: urls.api('index', 'v1', `task/${namespace}/artifacts/${name}`),
+      };
+    }
+
+    // We could use queue.buildUrl, but this creates URLs where the artifact
+    // name has slashes encoded. For artifacts we specifically allow slashes
+    // in the name unencoded, as this make things like `wget ${URL}` create
+    // files with nice names.
+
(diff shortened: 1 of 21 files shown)


helfi92:
screen shot 2018-08-02 at 4 12 40 pm

---

screen shot 2018-08-02 at 4 13 40 pm

src/App/index.jsx
```diff
@@ -41,6 +41,62 @@ import routes from './routes';
     '[disabled] .mdi-icon': {
       fill: theme.palette.primary.light,
     },
+    '.json-inspector__leaf': {
+      '&.json-inspector__leaf_root': {
```
eliperelman: lol

src/utils/buildArtifactUrl.js
```diff
@@ -0,0 +1,49 @@
+import { isNil } from 'ramda';
+import getIconFromMime from './getIconFromMime';
+import urls from './urls';
+
+export default ({ name, contentType, namespace, runId, taskId }) => {
```
eliperelman: I'm not sure we need a lot of this because I built into taskcluster-web-server that you get the artifact URLs automatically built from the server instead of having to do this on the client.

src/utils/buildArtifactUrl.js
```diff
@@ -0,0 +1,49 @@
...
+    }
+
+    return {
+      icon,
+      name,
+      url: urls.api('queue', 'v1', `task/${taskId}/artifacts/${name}`),
+    };
+  }
+
+  // TODO: If we have a userSession we should create a signed URL
```
eliperelman: Yeah, the server should do this for us automatically when logged in.

src/utils/buildArtifactUrl.js
helfi92: @eliperelman I didn't get a URL response from tc-web-server. The `url` field was `null`. Is this a bug in tc-web-server?

eliperelman: Possibly, but it should be a property of the artifact itself. Like `artifact.url`. Maybe try it out with the GraphQL playground.

helfi92:
screen shot 2018-09-19 at 4 12 37 pm
screen shot 2018-09-19 at 4 12 49 pm

PR #89, opened 2018-08-09
Add pulse messages view

helfi92:


src/views/PulseMessages/index.jsx
```diff
@@ -0,0 +1,243 @@
...
+              </Typography>
+            </Toolbar>
+            <DataTable
+              items={messages}
+              noItemsMessage="No messages received."
+            />
+          </List>
+          {listening ? (
+            <Tooltip title="Stop Listening">
+              <div className={classes.fab}>
```
eliperelman: This button should probably change color when listening starts, like red.

helfi92: Yep! It does exactly that via the stopIcon class:

```jsx
stopIcon: {
  ...theme.mixins.errorIcon,
},
```

eliperelman: Ah there it is, sweet.

helfi92:
@eliperelman I plan to merge this after https://github.com/taskcluster/taskcluster-web/pull/88 to remove the extra overhead on you.

PR #91, opened 2018-08-09
Add recent tasks to /tasks

### LINKED ISSUE #23: Recent tasks
When visiting the `/tasks` view without a task ID, let's display a list of recent tasks the user has viewed.

### LINKED ISSUE #87: Fix search bar in `/tasks`
Once a task is selected in `/tasks`, the search bar becomes no longer usable. This needs to be fixed.

helfi92:
Resolves #23.
Resolves #87.

### DIFF
--- a/src/utils/getTaskIdHistory.js
+++ b/src/utils/getTaskIdHistory.js
@@ -1,8 +1,0 @@
-import storage from 'localforage';
-import { RECENT_TASKS_STORAGE_KEY } from './constants';
-
-export default async () => {
-  const recentTasks = await storage.getItem(RECENT_TASKS_STORAGE_KEY);
-
-  return recentTasks || [];
-};
--- a/src/utils/getTaskIdHistory.js
+++ b/src/utils/getTaskIdHistory.js
@@ -1,5 +1,8 @@
+import storage from 'localforage';
 import { RECENT_TASKS_STORAGE_KEY } from './constants';
 
-export default JSON.parse(
-  localStorage.getItem(RECENT_TASKS_STORAGE_KEY) || '[]'
-);
+export default async () => {
+  const recentTasks = await storage.getItem(RECENT_TASKS_STORAGE_KEY);
+
+  return recentTasks || [];
+};
--- a/src/utils/getTaskIdHistory.js
+++ b/src/utils/getTaskIdHistory.js
@@ -1,0 +1,5 @@
+import { RECENT_TASKS_STORAGE_KEY } from './constants';
+
+export default JSON.parse(
+  localStorage.getItem(RECENT_TASKS_STORAGE_KEY) || '[]'
+);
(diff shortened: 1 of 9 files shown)


helfi92:
screen shot 2018-08-09 at 4 18 04 pm

src/utils/getTaskIdHistory.js
```diff
@@ -0,0 +1,5 @@
+import { RECENT_TASKS_STORAGE_KEY } from './constants';
+
+export default JSON.parse(
```
eliperelman: This is definitely something where you want to use [dexie](http://dexie.org/) as you can store these as strings in a table instead of needing to do the serialization dance.

src/views/Tasks/ViewTask/index.jsx
```diff
@@ -43,20 +69,35 @@ import pageArtifactsQuery from './pageArtifacts.graphql';
...
-          ? nextProps.match.params.taskId || ''
-          : prevState.taskSearch,
-      showError: !!nextProps.data.error,
+  constructor(props) {
+    super(props);
+
+    const taskId = props.match.params.taskId || '';
+
+    this.state = {
+      taskSearch: taskId,
```
eliperelman: This can also be done in `getDerivedStateFromProps`.

src/views/Tasks/ViewTask/index.jsx
```diff
@@ -43,20 +69,35 @@ import pageArtifactsQuery from './pageArtifacts.graphql';
...
-  state = {
-    taskSearch: '',
-    showError: false,
-  };
+  componentDidUpdate(prevProps) {
+    const { taskId } = this.props.match.params;
+
+    if (prevProps.match.params.taskId !== taskId) {
+      updateTaskIdHistory(taskId);
+      this.props.history.push(`/tasks/${taskId}`);
```
eliperelman: Not sure why this push is here; this should already be redirected to if `this.props.match.params` has changed.

src/utils/getTaskIdHistory.js
helfi92: Interesting. Do you prefer dexie over localforage? We should probably use only one indexDB library wrapper.

eliperelman: They each have their purpose. localforage is good for doing key-value stores like localstorage, where you just need to look up a single value. dexie is good for collections of data, like capturing a list.

helfi92:
@eliperelman If we go down that route, I believe we'll have to also limit the number of tasks in `db.recentTasks` otherwise, the db might max out especially when indexDB is not supported (fallsbacks to localstorage). From what I read, localforage is not good at multiple get/set operations in a batch but since we're just setting one array, it's fine. Let me know what you think.

eliperelman:
It's only one array for now, but I think user preferences and such will be eventually stored there, or maybe in the backend. Also, IDB is supported in all browsers that TC would support, so no fallback worries.

As far as limits, you can occasionally prune to the most recent count, but I don't see it growing massively immediately. 🤣 

Just my 2 cents.

PR #92, opened 2018-08-13
Add Create Task View

### LINKED ISSUE #24: Create task
This view is synonymous with https://tools.taskcluster.net/tasks/create. This also needs to take of creating tasks based on a task definition, synonymous with https://tools.taskcluster.net/tasks/JjZoye8zSSadEwUu_FPSkg/create. This does not concern interactive tasks. It may be nice to have a button on the page that can toggle between making the task interactive or not.

### LINKED ISSUE #25: Create interactive tasks
This view is synonymous with https://tools.taskcluster.net/tasks/JjZoye8zSSadEwUu_FPSkg/interactive. Basically any task ID following this route will do a look-up for its task definition, use that as the editor content, with overrides for interactive properties.

helfi92:
Resolves #24. 
Resolves #25.


helfi92:
screen shot 2018-08-13 at 11 57 35 am

package.json
```diff
@@ -58,6 +60,7 @@
     "react-loadable": "^5.3.1",
     "react-router-dom": "^4.3.1",
     "redbox-react": "^1.5.0",
+    "slugid": "^2.0.0",
```
eliperelman: Don't use this. Use the one from client-web or another web-compatible version.

eliperelman: Then you can remove Buffer.

helfi92: Oh that's true! Thanks for pointing that out 😍

PR #105, opened 2018-08-29
Add create/edit secret view

### LINKED ISSUE #47: Create/edit secret
This view is synonymous with https://tools.taskcluster.net/secrets/garbage%2Fcag%2Ftestsecret. It should be pre-filled when loading an existing secret, and empty when creating. I know right now we have a view-only mode for showing/hiding the secret value, but we can incorporate this into the form view, which should be the only view.

helfi92:
Resolves #47.


helfi92:
screen shot 2018-08-29 at 3 05 46 pm

PR #132, opened 2018-09-25
Issue#127Links in ErrorPanels are hard to read

preeti98:
Changed the color of the links to white to improve readability.

### DIFF
--- a/src/App/Main.jsx
+++ b/src/App/Main.jsx
@@ -28,8 +28,11 @@ import routes from './routes';
     '[disabled] .mdi-icon': {
       fill: theme.palette.primary.light,
     },
-    a: {
-      color: theme.palette.text.primary,
+   a: {
+      color: #FFF,
+    },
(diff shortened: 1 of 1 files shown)


helfi92:
Could you post a screenshot, please?

src/App/Main.jsx
```diff
@@ -28,8 +28,11 @@ import routes from './routes';
     '[disabled] .mdi-icon': {
       fill: theme.palette.primary.light,
     },
-    a: {
-      color: theme.palette.text.primary,
+   a: {
+      color: #FFF,
```
fiennyangeln: I think you need to wrap it as a string, something like `'#FFF'` to add CSS in JS

helfi92:
Fixed in https://github.com/taskcluster/taskcluster-web/commit/2dddb8209a591656767df6050c605e508f8449a4.

Issue #122, opened 2018-09-18
Add a Cache Purges entry to the navigation drawer


helfi92:
The navigation drawer doesn't seem to have an entry to Cache Purges. Let's add it.

![screen shot 2018-09-18 at 12 52 16 pm](https://user-images.githubusercontent.com/3766511/45703274-d6741400-bb41-11e8-87dc-8d9d2157eac1.png)

mike44441:
Hello! This is my first time working on something like this and would like to contribute.

helfi92:
Welcome @mike44441 :) The first step is to have your local setup running by going over the [README](https://github.com/taskcluster/taskcluster-web/blob/master/README.md). Once that's done, you'll be able to add a Cache Purges entry in the navigation drawer. Let me know if you get stuck somewhere.

mike44441:
Alright I have a question about picking the icon at the moment. Based on the other elements in the list, is there a specific one you would like or should I just place something as a filler for the time being in both the import and the list element?

helfi92:
Feel free to pick any icon for now. We can change it later once you send a pull-request :)

helfi92:
Completed in https://github.com/taskcluster/taskcluster-web/commit/0be756e920df625c36c14a81c07f151a9740d712.

PR #134, opened 2018-09-26
Issue #122 Adding Cache Purges Button to Navigation Drawer

mike44441:
Created a Cache Purges Button to Navigation Sidebar with unique icon.


.gitignore
```diff
@@ -7,3 +7,5 @@ node_modules
 .env
 /styleguide
 .coverage
+src/views/.DS_Store
```
helfi92: Replace with `.DS_Store`, instead of `src/views/.DS_Store`.

.gitignore
```diff
@@ -7,3 +7,5 @@ node_modules
...
+config.env
```
helfi92: What is this for?

mike44441: Ah that was an extra file I created. I'll remove everything that isn't the main modified file then.

mike44441:
I reverted my previous change and had the commit not change the .DS_Store and .gitignore files.

.gitignore
```diff
@@ -7,3 +7,4 @@ node_modules
 .env
 /styleguide
 .coverage
+.DS_Store
```
helfi92: I can still see `src/.DS_Store`, `src/components/.DS_Store` and `src/components/Dashboard/.DS_Store` commited:

screen shot 2018-09-27 at 8 22 55 am

You'll have to do something like:

```bash
git rm src/.DS_Store
git rm src/components/.DS_Store
git rm src/components/Dashboard/.DS_Store
git commit -m "remove DS_Store"
```

Then the last step would be to `git push` to this branch.

mike44441: Thanks for the help and I have just done finished the push. =)

mike44441:
I have pushed the changes, let me know if there is anything else!

Issue #56, opened 2018-04-10
App bar view help


eliperelman:
Each page could contain important information for the new user of a particular view, but often doesn't warrant needing to be shown every time. Let's put a help icon in the app bar to show help, perhaps in a dialog or right drawer, showing help for a page.

PR #149, opened 2018-10-05
Add view help section prop to the Dashboard component

helfi92:
Relates to #56.


helfi92:
screen shot 2018-10-05 at 2 14 01 pm

PR #157, opened 2018-10-10
Migrate tc-web to neutrino v9

helfi92:


.neutrinorc.js
```diff
@@ -1,37 +1,127 @@
...
-        Wrapper: join(__dirname, 'src/styleguide/ThemeWrapper.jsx'),
-        StyleGuideRenderer: join(__dirname, 'src/styleguide/StyleGuideRenderer.jsx'),
+    ['@neutrinojs/airbnb', {
+      eslint: {
+        parserOptions: {
+          ecmaFeatures: {
+            legacyDecorators: true
+          }
+        },
+        emitWarning: process.env.NODE_ENV === 'development',
```
edmorley: We'll likely be making this change upstream soon, see:
https://github.com/neutrinojs/neutrino/issues/1131#issuecomment-428982103

.neutrinorc.js
```diff
@@ -52,20 +142,19 @@ module.exports = {
...
     }],
-    ['@neutrinojs/env', [
-      'NODE_ENV',
-      'APPLICATION_NAME',
-      'LOGIN_STRATEGIES',
-      'TASKCLUSTER_ROOT_URL',
-    ]],
     (neutrino) => {
       if (process.env.NODE_ENV === 'development') {
         neutrino.config.devtool('cheap-module-source-map');
```
edmorley: I'd move this to the `devtool` option within `@neutrinojs/react`, since it ensures terser-webpack-plugin will be configured correctly (switching to terser is currently unreleased; see https://github.com/neutrinojs/neutrino/pull/1158).

Also, the default is `cheap-module-eval-source-map`, which is likely faster for incremental builds (given eval), so unless there is some issue being encountered, might be best left at the default. See:
https://webpack.js.org/configuration/devtool/#devtool

.neutrinorc.js
```diff
@@ -79,6 +168,10 @@ module.exports = {
           bindings: 'bindings'
         }));
 
+      // TODO: Why is this needed?
+      neutrino.config.resolve.modules.add(MODULES);
```
edmorley: I would recommend not adding this. If not having it causes an error, could you include it in the code comment? (And I'm happy to take a look)

.neutrinorc.js
```diff
@@ -96,6 +189,8 @@ module.exports = {
           .test(/JSONStream/)
           .use('shebang')
             .loader('shebang-loader');
+
+      console.log(JSON.stringify(neutrino.config.toConfig(), null, 2));
```
edmorley: Manually stringifying the config is not neccessary/gives worse output. Try:
````
neutrino --inspect --mode {development,production}
# or
NODE_ENV={development,production,test} neutrino --inspect
```

(https://master.neutrinojs.org/usage/#inspecting-the-generated-webpack-config)

karma.conf.js
```diff
@@ -0,0 +1,5 @@
+const neutrino = require('neutrino');
+
+process.env.NODE_ENV = process.env.NODE_ENV || 'test';
+
+module.exports = neutrino().karma();
```
edmorley: (Missing end of file newline)

.neutrinorc.js
```diff
@@ -52,20 +142,19 @@ module.exports = {
...
         },
       },
-      eslint: {
-        rules: {
-          // This has been set in the latest Airbnb preset, but has not been
-          // released yet.
-          'react/no-did-mount-set-state': 'off',
-        }
-      }
+      env: [
```
edmorley: If it helps, this `env` option now also supports the object-form that can take defaults for each variable.

.neutrinorc.js
```diff
@@ -79,6 +168,10 @@ module.exports = {
           bindings: 'bindings'
         }));
 
+      // TODO: Why is this needed?
+      neutrino.config.resolve.modules.add(MODULES);
+      // neutrino.config.resolveLoader.modules.add(MODULES);
+
       neutrino.config.output.publicPath('/');
```
edmorley: This can be set via the `@neutrinojs/react` options now, which will also ensure the devserver equivalent option is set to a matching value automatically.

.neutrinorc.js
edmorley: It looks like the globally `yarn link`ed version of Neutrino is using an older version of webpack-chain. I would either:
1. run `yarn neutrino ...` (in the hope that it favours the copy of Neutrino in the local `node_modules`)
2. remove the yarn link for neutrino
3. update the `neutrino-dev` source directory to `master` and run `yarn` to install the latest version of deps

(I would suggest 1/2, otherwise we'll be debugging Neutrino `master` rather than the known release version, which is not what `yarn start` will be using)

helfi92: Created https://github.com/neutrinojs/neutrino/issues/1167.

helfi92: I'm not sure I am running a linked version of Neutrino. Running `grep "webpack-chain" node_modules/neutrino/package.json` shows `"webpack-chain": "^4.11.0",`. The update to webpack-chain was merged [two days ago](https://github.com/neutrinojs/neutrino/commit/84bdf6a08383bfd69d7f3c93f0e48117893ce84e). Is it possible that it hasn't been published yet?

edmorley: If `neutrino` is on `PATH` (ie `neutrino` works and not just `yarn neutrino`), it's a linked copy, or else `PATH` has been manually adjusted.

That grep command is a red herring, since that will be grepping the local node_modules copy, not the copy that's being used from the linked neutrino.

Try running this:
`ls -al $(yarn global bin)/neutrino`

helfi92: Removing the linked version of Neutrino works! Many thanks!

edmorley: Awesome :-)

helfi92:
Blocked by https://github.com/neutrinojs/neutrino/issues/1166.

package.json
```diff
@@ -4,22 +4,20 @@
...
-    "start": "neutrino start --require dotenv/config",
-    "test": "neutrino test",
-    "lint": "neutrino lint",
-    "precommit": "neutrino stage",
-    "create:fragment-matcher": "node src/scripts/queryServer",
-    "styleguide": "neutrino styleguide:start --require dotenv/config"
+    "build": "yarn create:fragment-matcher && dotenv webpack -- --mode production",
+    "start": "dotenv webpack-dev-server -- --mode development",
+    "test": "karma start --single-run",
+    "lint": "eslint --cache --format codeframe --ext mjs,js src test",
```
edmorley: This needs to be `--ext mjs,jsx,js`

helfi92: Wow, that was silly from my end. Sorry for that.

edmorley: I didn't see it the first time I looked either 😆

helfi92:
OK, I think this is ready. I'll merge after https://github.com/taskcluster/taskcluster-web/pull/160 if there are no objections.

/cc @djmitche.

djmitche:
None from me!

PR #160, opened 2018-10-12
Add client view initial search content #158

### LINKED ISSUE #158: Client view initial search content
When logged in, navigating to the clients view should have the search bar should have its value set as the current user client ID.

screen shot 2018-10-11 at 7 58 40 am

For example, if the current logged in user has `mozilla-auth0/ad|Mozilla-LDAP|haali` as their client ID then the search bar should have this value as well as perform a search in order to filter the table with proper values. It's perfectly fine if the table becomes empty since our database doesn't always return a value for paginated results. We have a separate bug for that https://github.com/taskcluster/taskcluster-web/issues/67.

screen shot 2018-10-11 at 8 06 50 am

projectyang:
Set client id as default search in Authorization/Clients

Added client id as default state in Search component along with initial search query for the specified id on page load.

Fixes #158, however users are now unable to retrieve a 'full' default clients view (e.g. expected search query with '' to retrieve full list of clients does not work)

### DIFF
--- a/src/views/Clients/ViewClients/index.jsx
+++ b/src/views/Clients/ViewClients/index.jsx
@@ -13,8 +13,10 @@ import Button from '../../../components/Button';
 import ClientsTable from '../../../components/ClientsTable';
 import { VIEW_CLIENTS_PAGE_SIZE } from '../../../utils/constants';
 import clientsQuery from './clients.graphql';
+import { withAuth } from '../../../utils/Auth';
 
 @hot(module)
+@withAuth
 @graphql(clientsQuery, {
   options: () => ({
     variables: {
@@ -30,8 +32,11 @@ import clientsQuery from './clients.graphql';
   },
 }))
 export default class ViewWorker extends PureComponent {
+  componentDidMount() {
+    this.handleClientSearchSubmit();
+  }
   state = {
-    clientSearch: '',
+    clientSearch: this.props.user.credentials.clientId,
   };
 
   handlePageChange = ({ cursor, previousCursor }) => {
@@ -78,7 +83,9 @@ export default class ViewWorker extends PureComponent {
   };
 
   handleClientSearchSubmit = e => {
-    e.preventDefault();
+    if (e) {
+      e.preventDefault();
+    }
 
     const {
       data: { refetch },


src/views/Clients/ViewClients/index.jsx
```diff
@@ -30,8 +32,11 @@ import clientsQuery from './clients.graphql';
   },
 }))
 export default class ViewWorker extends PureComponent {
+  componentDidMount() {
+    this.handleClientSearchSubmit();
+  }
   state = {
-    clientSearch: '',
+    clientSearch: this.props.user.credentials.clientId,
```
helfi92: If you log out and then refresh the page, the view will crash on this line. This is because `this.props.user` is undefined.

You can do:

```jsx
clientSearch: this.props.user ? this.props.user.credentials.clientId || '',
```

projectyang:
I also noticed that if I were to query with spaces '     ', the following error appears:
screen shot 2018-10-12 at 3 36 08 pm

Should we attempt to sanitize any leading spaces in the search value before querying?

helfi92:
Good catch. Sanitization would be great to have!

src/views/Clients/ViewClients/index.jsx
```diff
@@ -82,11 +82,16 @@ export default class ViewWorker extends PureComponent {
...
+  handleClientSearchSubmit = async e => {
     if (e) {
       e.preventDefault();
     }
 
+    await this.setState({
+      clientSearch: this.state.clientSearch.replace(/\s/g, ''),
+    });
+    console.log(`${this.state.clientSearch}test`)
+
```
helfi92: No need to modify the state. When refetching you can simply remove trailing spaces from both ends using the `trim` method:

```jsx
clientOptions: {
  ...(clientSearch ? { prefix: clientSearch.trim() } : null),
},
``

File: src/views/Clients/ViewClients/index.jsx [hunk 3]
```diff
@@ -82,11 +82,16 @@ export default class ViewWorker extends PureComponent {
     this.setState({ clientSearch: target.value });
   };
 
-  handleClientSearchSubmit = e => {
+  handleClientSearchSubmit = async e => {
     if (e) {
       e.preventDefault();
     }
 
+    await this.setState({
```
helfi92: Remove this line.

src/views/Clients/ViewClients/index.jsx
```diff
@@ -82,11 +82,16 @@ export default class ViewWorker extends PureComponent {
     this.setState({ clientSearch: target.value });
   };
 
-  handleClientSearchSubmit = e => {
+  handleClientSearchSubmit = async e => {
```
helfi92: Remove `async`.

projectyang:
Thanks for the help! Really appreciate the time you took to explain everything.

src/views/Clients/ViewClients/index.jsx
```diff
@@ -30,8 +32,11 @@ import clientsQuery from './clients.graphql';
   },
 }))
 export default class ViewWorker extends PureComponent {
+  componentDidMount() {
+    this.handleClientSearchSubmit();
+  }
```
helfi92: This is going to fetch the list of clients twice. Let's rely on the initial graphql call to perform search the filtered clients. You can achieve that by adding the same logic for `clientsOptions` under `variables`. See other comment.

src/views/Clients/ViewClients/index.jsx
```diff
@@ -13,8 +13,10 @@ import Button from '../../../components/Button';
 import ClientsTable from '../../../components/ClientsTable';
 import { VIEW_CLIENTS_PAGE_SIZE } from '../../../utils/constants';
 import clientsQuery from './clients.graphql';
+import { withAuth } from '../../../utils/Auth';
 
 @hot(module)
+@withAuth
 @graphql(clientsQuery, {
   options: () => ({
```
helfi92: `options` can have access to props:

```jsx
options: (props) => ({
```

src/views/Clients/ViewClients/index.jsx
```diff
@@ -13,8 +13,10 @@ import Button from '../../../components/Button';
...
     variables: {
```
helfi92: We can remove `componentDidMount` by including some logic here:

```jsx
clientOptions: {
  // add similar logic as `refetch` in `handleClientSearchSubmit`.
},
```

src/views/Clients/ViewClients/index.jsx
```diff
@@ -78,21 +83,19 @@ export default class ViewWorker extends PureComponent {
   };
 
   handleClientSearchSubmit = e => {
-    e.preventDefault();
+    if (e) {
+      e.preventDefault();
+    }
```
helfi92: Since we won't be calling this method from `componentDidMount` anymore, we can remove the if statement and just use `e.preventDefault()`.

projectyang:
I've made the recommendations, but discovered two scenarios that may cause issues for users:

Scenario 1:
If a user logs in or out on the Clients view, the search bar does not update automatically. This can lead to confusion if a user logs in (and triggers the default clientId query) and does not know about it.

screen shot 2018-10-15 at 10 59 22 am

My proposed solution is to use ComponentDidUpdate in the clients component to monitor to detect when a log in/out has occurred and update state to match.

Scenario 2:
Related to the above, if a user logins in our out on the Clients view and error is thrown by Apollo:

screen shot 2018-10-15 at 10 36 56 am

I'm having difficulty solving this one. My assumption is that the graphql query for the Clients component is still in flight when resetStore() is called, thus throwing the error. Is the correct approach to wait for the query to finish before triggering resetStore()?

helfi92:
> I'm having difficulty solving this one. My assumption is that the graphql query for the Clients component is still in flight when resetStore() is called, thus throwing the error.

See if the following fixes it:

```diff
diff --git a/src/components/Dashboard/UserMenu.jsx b/src/components/Dashboard/UserMenu.jsx
index 3ff85f3..d573cab 100644
--- a/src/components/Dashboard/UserMenu.jsx
+++ b/src/components/Dashboard/UserMenu.jsx
@@ -59,12 +59,12 @@ export default class UserMenu extends Component {
     this.setState({ signInDialogOpen: false });
   };
 
-  handleClickSignOut = () => {
+  handleClickSignOut = async () => {
     this.handleMenuClose();
-    this.props.onUnauthorize();
     // Since Apollo caches query results, it’s important to get rid of them
     // when the login state changes.
-    this.props.client.resetStore();
+    await this.props.client.resetStore();
+    this.props.onUnauthorize();
   };
 
   render() {
diff --git a/src/components/SignInDialog/index.jsx b/src/components/SignInDialog/index.jsx
index 80983fd..b03c6e9 100644
--- a/src/components/SignInDialog/index.jsx
+++ b/src/components/SignInDialog/index.jsx
@@ -44,11 +44,14 @@ export default class SignInDialog extends Component {
     );
   }
 
-  handleCredentialsSignIn = credentials => {
+  handleCredentialsSignIn = async credentials => {
     const inOneWeek = new Date();
 
     inOneWeek.setDate(inOneWeek.getDate() + 7);
 
+    // Since Apollo caches query results, it’s important to get rid of them
+    // when the login state changes.
+    await this.props.client.resetStore();
     this.props.onAuthorize({
       credentials,
       expires: inOneWeek.toISOString(),
@@ -57,9 +60,6 @@ export default class SignInDialog extends Component {
         displayName: credentials.clientId,
       },
     });
-    // Since Apollo caches query results, it’s important to get rid of them
-    // when the login state changes.
-    this.props.client.resetStore();
     this.props.onClose();
   };
```

projectyang:
That fixed it!

I ended up using [getDerivedStateFromProps](https://reactjs.org/blog/2018/03/27/update-on-async-rendering.html#updating-state-based-on-props) to solve scenario one. Let me know your thoughts on that approach. 

The eslint error is appearing since I only reference a state field inside getDerivedStateFromProps. Here's a [relevant thread](https://github.com/yannickcr/eslint-plugin-react/issues/1759) with a PR at the end that addresses this.

src/views/Clients/ViewClients/index.jsx
```diff
@@ -31,9 +36,28 @@ import clientsQuery from './clients.graphql';
 }))
 export default class ViewWorker extends PureComponent {
   state = {
-    clientSearch: '',
+    clientSearch: this.props.user ? this.props.user.credentials.clientId : '',
```
helfi92: Could you rename the class from `ViewWorker` to `ViewClients`, please? It must have been a typo from the past :)

src/views/Clients/ViewClients/index.jsx
```diff
@@ -31,9 +36,28 @@ import clientsQuery from './clients.graphql';
...
+    clientId: this.props.user ? this.props.user.credentials.clientId : '',
```
helfi92: Given that we're not using `getDerivedStateFromProps` which will run when the site is first loaded, we can do:

```jsx
state = {
  clientSearch: '',
  previousClientId: '',
};
```

helfi92: Rename to `clientId` to `previousClientId` otherwise it can be difficult to differentiate between `state.clientId` and `props.users.credentials.clientId`.

helfi92:
Let's make Travis happy :)

https://travis-ci.org/taskcluster/taskcluster-web/builds/442345832?utm_source=github_status&utm_medium=notification

projectyang:
Travis is hard to please...For the error it's failing on:

_error: Unused state field: 'previousClientId' (react/no-unused-state) at src/views/Clients/ViewClients/index.jsx:40:5:_

I believe it should pass since we are referring to the state previousClientId in getDerivedStateFromProps. This seems to be more of a problem with the eslint rules than the code in my opinion. 

See this [thread here](https://github.com/yannickcr/eslint-plugin-react/pull/1829) - which reflects updates made to eslint-plugin-react which allows for this to pass. 

The other option could be trying to incorporate previousClientId somewhere else in the component, which seems unnecessary if it's just to satisfy a rule.

projectyang:
@eliperelman, can you help clarify for me why this is a smell? Apologies for my inexperience with React.

I was under the impression that this pattern was acceptable based on the [React docs example](https://reactjs.org/blog/2018/03/27/update-on-async-rendering.html#fetching-external-data-when-props-change).

My belief that this an issue with the eslint package is based on this [thread here](https://github.com/yannickcr/eslint-plugin-react/issues/1759) for eslint-plugin-react where others have complained about similar situations regarding no-unused-state and getDerivedStateFromProps, which was eventually resolved and merged into master [here](https://github.com/yannickcr/eslint-plugin-react/pull/1829).

src/views/Clients/ViewClients/index.jsx
```diff
@@ -29,11 +34,33 @@ import clientsQuery from './clients.graphql';
     ...theme.mixins.fab,
   },
 }))
-export default class ViewWorker extends PureComponent {
+export default class ViewClients extends PureComponent {
   state = {
     clientSearch: '',
+    previousClientId: '',
```
helfi92: You can remove the error with:

```jsx
// eslint-disable-next-line react/no-unused-state
previousClientId: '',
```

It's also done in https://github.com/taskcluster/taskcluster-web/blob/04d3822572f7bbb6ae831ad6edbe9e5ad2c61898/src/components/ClientForm/index.jsx#L99.

PR #167, opened 2018-10-15
Add page and event analytics

### LINKED ISSUE #51: Page analytics
Every view navigation should register an analytics event.

### LINKED ISSUE #52: Action analytics
Every interaction, like button clicks, should register an analytics event. We will probably have to encapsulate this our own button components.

helfi92:
Fixes #52.
Fixes #51.

### DIFF
--- a/src/components/RouteWithProps/index.jsx
+++ b/src/components/RouteWithProps/index.jsx
@@ -1,10 +1,12 @@
 import React, { Component } from 'react';
 import { Route } from 'react-router-dom';
+import withPageTracker from '../../utils/withPageTracker';
 
 /**
  * Conditionally render a component based on location, with non-react-router
  * specific properties forwarded to the rendering component.
  */
+@withPageTracker
 export default class RouteWithProps extends Component {
   render() {
     const {
@@ -24,7 +26,9 @@ export default class RouteWithProps extends Component {
         strict={strict}
         location={location}
         sensitive={sensitive}
-        render={renderProps => <Component {...renderProps} {...props} />}
+        render={({ staticContext, ...renderProps }) => (
+          <Component {...renderProps} {...props} />
+        )}
       />
     );
   }
--- a/src/utils/prop-types.js
+++ b/src/utils/prop-types.js
@@ -360,16 +360,17 @@ export const pulseMessage = shape({
 // https://developers.google.com/analytics/devguides/collection/analyticsjs/events
 export const gaEvent = shape({
   /**
-   * A description of the behaviour.
-   * E.g. 'Clicked Delete', 'Added a component', 'Deleted account', etc.
-   * */
-  label: string,
-  /**
    * Defaults to 'Click'.
    * A description of the behaviour.
    * E.g. 'Clicked Delete', 'Added a component', 'Deleted account', etc.
    * */
   action: string,
+  /**
+   * More precise labelling of the related action.
+   * E.g. alongside the 'Added a component' action, we could add the name
+   * of a component as the label. E.g. 'Survey', 'Heading', 'Button', etc.
+   * */
+  label: string,
   /**
    * Defaults to 'Uncategorized'.
    * A top level category for the event.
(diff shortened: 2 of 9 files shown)


helfi92:
Agreed. I'll send a patch for that.

helfi92:
Done.

src/utils/withPageTracker.jsx
```diff
@@ -0,0 +1,26 @@
+import React, { Component } from 'react';
+import ReactGA from 'react-ga';
+
+let currentPage = null;
```
eliperelman: I think this is a bad idea as this could be mutated from anywhere. I would instead make this track based on the decorator if you need to override:

```js
@withPageTracker("/tasks")
```

helfi92: Probably a silly question, but how could it be mutated if it's not exported?

eliperelman: You are exporting the component which exposes `trackPage`, which mutates a variable that doesn't change with each export, so every component that calls `trackPage` would all mutate the same `currentPage`.

eliperelman: Worst case scenario, you can store the `currentPage` on `this`:

```js
currentPage = this;

trackPage(page) {
  this.currentPage = page;
}
```

src/utils/prop-types.js
```diff
@@ -348,3 +348,41 @@ export const cachePurge = shape({
...
+export const gaEvent = shape({
+  /**
+   * A description of the behaviour.
+   * E.g. 'Clicked Delete', 'Added a component', 'Deleted account', etc.
+   * */
+  label: string,
+  /**
+   * Defaults to 'Click'.
+   * A description of the behaviour.
+   * E.g. 'Clicked Delete', 'Added a component', 'Deleted account', etc.
```
djmitche: different E.g.'s here?

src/utils/withPageTracker.jsx
djmitche: Aside from the use of the odious `=`, that seems like the best case. :)

eliperelman: It's a class property lol.

helfi92: > Worst case scenario, you can store the currentPage on this.

I like this approach. Thanks.

helfi92:
Rebased.

src/components/RouteWithProps/index.jsx
```diff
@@ -24,7 +26,9 @@ export default class RouteWithProps extends Component {
         strict={strict}
         location={location}
         sensitive={sensitive}
-        render={renderProps => <Component {...renderProps} {...props} />}
+        render={({ staticContext, ...renderProps }) => (
```
djmitche: I missed this on the first round -- what is `staticContext` and why does it need to be filtered out here?

src/utils/withPageTracker.jsx
```diff
@@ -0,0 +1,26 @@
+import React, { Component } from 'react';
+import ReactGA from 'react-ga';
+
+export default WrappedComponent =>
+  class WithPageTracker extends Component {
+    currentPage = null;
```
djmitche: I may not understand things here, but even as a class property, there will be one class per page, right?  So this will start out `null`, then be set to the location when the user navigates to this page, and then never be reset again?

For example:
 * user clicks to page A; A.currentPage = "/page/a"; ReactGA.pageview("/page/a");
 * user clicks to page B; B.currentPage = "/page/b"; ReactGA.pageview("/page/b");
 * user clicks to page A; A.currentPage is already "/page/a" so nothing happens

src/components/RouteWithProps/index.jsx
helfi92: A `staticContext` property is inserted whenever we use react-router. If we don't remove it, it will end up as a prop to another compoent that does not recognize this property and an error will be thrown in the console: ``` "React does not recognize the `staticContext` prop on a DOM element"  ```.

djmitche: Ah, OK :)

Issue #163, opened 2018-10-14
Add copy to clipboard feature


helfi92:
Steps to reproduce:

1. Navigate to http://localhost:5080/tasks/Aia7BIyvRGexdGaoeqT8NA
2. Click on the "Created" row

<img width="1146" alt="screen shot 2018-10-14 at 11 48 01 am" src="https://user-images.githubusercontent.com/3766511/46919040-172b4580-cfa7-11e8-9add-864d5c2fb81c.png">

Expected: The date time is copied to the clipboard
Current: The date time not copied to the clipboard

Mryashbhardwaj:
> http://localhost:5080/tasks/Aia7BIyvRGexdGaoeqT8NA

the link here is specific to the tasks that you have added in you local , right ?
for me to be able to replicate , would  i have to add my own tasks ?

helfi92:
You should be able to view this task without having to create your own task. If you're not able to view the task, let me know what kind of error you are seeing and I'll try to help.

Mryashbhardwaj:
https://github.com/taskcluster/taskcluster-web/issues/162#issuecomment-429630548

this is the error i see when i try to open that link . 
Note : i am logged in using some random values when this error came .
the client taskcluster-web and taskcluster-web-server are able to comunicate as i can see 
> https://paste.fedoraproject.org/paste/PsftxFEzycg9hwzFhXw1jQ
on the server terminal.

also i get  a diffrent erro when i sign out .

![2018-10-15-010915_1209x750_scrot](https://user-images.githubusercontent.com/13954513/46921283-277b0900-d017-11e8-83c0-b29dd6ca9aea.png)

server terminal output:
> https://paste.fedoraproject.org/paste/wcOUWfRedcqx~BQb7yiu4A

also , following is my .env file 

> APPLICATION_NAME="Taskcluster"
> AUTH0_DOMAIN="auth.mozilla.auth0.com"
> AUTH0_CLIENT_ID="29t2n3LKKnyTbGtWmfTkQpau0mp7QmMH"
> AUTH0_REDIRECT_URI="http://localhost:5080/login"
> AUTH0_RESPONSE_TYPE="token id_token"
> AUTH0_SCOPE="openid profile"
> TASKCLUSTER_ROOT_URL="http://localhost:3050"
> LOGIN_STRATEGIES=""
> PORT="5080"

helfi92:
[...]

PR #172, opened 2018-10-18
Add copy to clipboard feature #163

Mryashbhardwaj:
Used [nkbt/react-copy-to-clipboard](https://github.com/nkbt/react-copy-to-clipboard)  for copying to the clipboard. I am able to copy the raw data 

> Example :
2018-10-15T12:07:34.946Z

 successfully.

### DIFF
--- a/src/components/TaskDetailsCard/index.jsx
+++ b/src/components/TaskDetailsCard/index.jsx
@@ -1,6 +1,7 @@
 import { Component, Fragment } from 'react';
 import { arrayOf, shape, string } from 'prop-types';
 import { Link } from 'react-router-dom';
+import { CopyToClipboard } from 'react-copy-to-clipboard';
 import Code from '@mozilla-frontend-infra/components/Code';
 import Label from '@mozilla-frontend-infra/components/Label';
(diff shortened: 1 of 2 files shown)


src/components/TaskDetailsCard/index.jsx
```diff
@@ -143,7 +144,9 @@ export default class TaskDetailsCard extends Component {
                   primary="Created"
                   secondary={<DateDistance from={task.created} />}
                 />
-                <ContentCopyIcon />
+                <CopyToClipboard text={task.created}>
+                  <ContentCopyIcon />
+                </CopyToClipboard>
```
helfi92: We want the copy action to happen not only when we click the copy icon but also when we click anywhere in the list item. That said, I think we can do:

```jsx
<CopyToClipboard text={task.created}>
  <ListItem button className={classes.listItemButton}>
    <ListItemText
      primary="Created"
      secondary={<DateDistance from={task.created} />}
    />
    <ContentCopyIcon />
  </ListItem>
</CopyToClipboard>
```

PR #173, opened 2018-10-18
Add styleguide

helfi92:


package.json
```diff
@@ -11,6 +11,8 @@
     "build": "yarn create:fragment-matcher && dotenv webpack -- --mode production",
     "start": "dotenv webpack-dev-server -- --mode development",
     "test": "karma start --single-run",
+    "styleguide": "styleguidist server --config styleguide.config.js",
```
eliperelman: The `--config styleguide.config.js` is the default so is unnecessary.

styleguidist.js
```diff
@@ -0,0 +1,14 @@
+const { join } = require("path");
```
eliperelman: You probably don't need this in a separate file and can be inlined into the function format of the neutrinorc.

helfi92:
Thank you @eliperelman for your input as always!

PR #177, opened 2018-10-19
Add copy to clipboard feature #163 , extended to all places where applicable

Mryashbhardwaj:
i have extended the feature to all the places where "ContentCopyIcon" was used.
Note: in cases when the icon was in a table cell, i have make the whole cell to triger a copy to clipboard.

Also in cases where ternary conditions were used to displlay content  as in 
```

run.resolved ? (
                        <DateDistance
                          from={run.resolved}
                          offset={run.started}
                        />
                      ) : (
                        <em>n/a</em>
                      )
```

i have used
`<CopyToClipboard text={run.resolved}>
`
directly because in such cases if the condition variabe is null/""  nothing will be coppied to the clipboard.
Note :  Following components have not been tested.
> ./src/components/WorkerTable/index.jsx
./src/components/WorkerTypesTable/index.jsx
./src/components/WorkersTable/index.jsx

Let me know if that is fine.


helfi92:
🎉

PR #182, opened 2018-10-20
Persist dark/light theme

### LINKED ISSUE #174: Persist dark/light theme
Steps to reproduce:

1. Navigate to http://localhost:5080/provisioners. The theme of the site will be set to "dark"
2. Click the light bulb icon in the top app bar. The theme is now light.
3. Refresh the page

Expected: The refreshed page has a light theme
Actual: The refreshed page has a dark theme

To implement this, you can use `localStorage` to help persist data.

vaishnavipamulapati:
Fixes #174.

### DIFF
--- a/src/App/index.jsx
+++ b/src/App/index.jsx
@@ -21,6 +21,7 @@ import FontStager from '../components/FontStager';
 import Main from './Main';
 import { ToggleThemeContext } from '../utils/ToggleTheme';
 import { AuthContext } from '../utils/Auth';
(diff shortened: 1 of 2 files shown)


helfi92:
@eliperelman Making an asynchronous call in `componentDidMount` will make the component render twice. Is there any way we can prevent that? ? A quick solution would be to use localStorage since it has a synchronous API but that would mean we'll be slowing down the first render.

PR #192, opened 2018-10-23
Create customized snackbar component

### LINKED ISSUE #178: Create a customized snackbar component
We will need to use a [customized snackbar](https://material-ui.com/demos/snackbars/#customized-snackbars) component to give feedback to users. Let's create a component in `src/component/Snackbar` capable of doing that.

The props of the component:

* `variant`: One of `error`, `warning`, `info`, and `success`. The variant will affect the background color of the Snackbar as well as add an icon from [`mdi-react`](https://materialdesignicons.com/).
* `message`: `string`.

Any other prop fed to `src/components/Snackbar` should be fed to material-ui's Snackbar component.

For inspiration, you can look at https://github.com/taskcluster/taskcluster-web/blob/master/src/components/Button/index.jsx.

helfi92:
Fixes #178.

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,7 +11,7 @@
     "build": "dotenv webpack -- --mode production",
     "start": "dotenv webpack-dev-server -- --mode development",
     "test": "karma start --single-run",
-    "styleguide": "styleguidist server",
+    "styleguide": "dotenv styleguidist server",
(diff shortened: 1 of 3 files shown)


helfi92:
screen shot 2018-10-23 at 10 36 52 am

PR #202, opened 2018-10-23
Ability to close all error panels #118

### LINKED ISSUE #118: Add ability to close all error panels
Let's give the user the ability to close all error panels in the site.

screen shot 2018-09-17 at 7 57 05 pm

projectyang:
Add ability for user to close error panels
Fixes #118

I'm not confident about my approach here using two state variables, it seems a bit inelegant. However, I can't think of another way to simulate error in the component state instead of props. Any suggestions would be appreciated.


src/views/Clients/ViewClients/index.jsx
```diff
@@ -201,4 +228,4 @@ export default class ViewClients extends PureComponent {
       </Dashboard>
     );
   }
-}
+}
```
helfi92: `handleClientSearchSubmit` seems to be defined two times. Remove the old implementation.

src/views/Clients/ViewClients/index.jsx
```diff
@@ -39,6 +39,10 @@ export default class ViewClients extends PureComponent {
     clientSearch: '',
     // eslint-disable-next-line react/no-unused-state
     previousClientId: '',
+    // eslint-disable-next-line react/no-unused-state
+    closeError: true,
+    // eslint-disable-next-line react/no-unused-state
+    error: false,
```
helfi92: Instead of having to implement this kind of logic in every component, what do you think about creating a new component `components/ErrorPanel` that would take care closing errors and re-displaying the error if a new one is given. Something like the following (not tested):

```jsx
import React, { Component } from 'react';
import { oneOfType, string, object } from 'prop-types';
import MuiErrorPanel from '@mozilla-frontend-infra/components/ErrorPanel';

export default class ErrorPanel extends Component {
  static propTypes = {
    /** Error to display. */
    error: oneOfType([string, object]),
  };

  static defaultProps = {
    error: null,
  };

  state = {
    error: null,
    // eslint-disable-next-line react/no-unused-state
    previousError: null,
  };

  static getDerivedStateFromProps(props, state) {
    if (props.error !== state.previousError) {
      return {
        error: props.error,
        previousError: props.error,
      };
    }

    return null;
  }

  handleErrorClose = () => {
    const { onClose } = this.props;

    this.setState({ error: null });

    if (onClose) {
      onClose();
    }
  };

  render() {
    const { error: _, onClose, ...props } = this.props;
    const { error } = this.state;

    return (
      error && (
        <MuiErrorPanel
          error={error}
          onClose={this.handleErrorClose}
          {...props}
        />
      )
    );
  }
}
```

projectyang:
Your code worked perfectly. Tested the new ErrorPanel component by removing await from resetStore() to trigger an error from logging in/out on the ViewClients view.

Question on a couple patterns I noticed when replacing the old ErrorPanel components:

`error && error.graphQLErrors && <ErrorPanel error={error.graphQLErrors[0].message} />`
vs
`data && data.error && data.error.graphQLErrors && <ErrorPanel error={data.error} />`
vs
`error && error.graphQLErrors && <ErrorPanel error={error} />`

1. Is there a reason we want specifically graphQLErrors[0].message for some views?
2. Is it necessary to do a check for data? What would be a case where it's undefined?
3. Is it necessary to do a check for a error.graphQLErrors? I assume that as long as error is present, graphQLErrors is also present.

PR #220, opened 2018-10-25
Update README with tips for running on non-localhost

djmitche:
I know this is probably mostly just me :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,7 +26,6 @@ To get started local development, create a file in the root of the repo named
 
 ```bash
 APPLICATION_NAME="Taskcluster"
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -26,7 +26,6 @@ To get started local development, create a file in the root of the repo named
 
 ```bash
 APPLICATION_NAME="Taskcluster"
-GRAPHQL_SUBSCRIPTION_ENDPOINT="ws://localhost:5080/subscription"
```
djmitche: I removed this because I think it's the default in `.neutrinorc`.

PR #222, opened 2018-10-25
Fix: Pagination in clients view

### LINKED ISSUE #213: Pagination on clients browser doesn't work
STR:
* https://taskcluster-web.netlify.com/auth/clients
* search for `project/taskcluster`
  * side note: there's no spinner to say that the search is running or completed here
* see 12 clients (not even a screenful..)
* click >
* see spinner, spinner resolves, same 12 clients

helfi92:
Resolves #213.


helfi92:
> Is the UI getting a blank page of results? That will make for a weird user experience, too, right?

Yes, the user will be presented with a blank page but note that this scenario will only happen when there are actually no more pages. I think this is a better experience than having to see nothing change after clicking the `>`. Clicking back would be almost instantaneous since the result of the previous page would have been cached.

Issue #226, opened 2018-10-26
Signing out not always working


helfi92:
STR:

1. Navigate to http://localhost:5080/aws-provisioner/gecko-t-win7-32 while being logged in
2. Sign out

Expected: The user is signed out and the text has been replace with "Sign In"
Actual: The user is not signed out

It appears like when the graphql response returns no result or an error, signing out is not working.

alecclarke:
@helfi92 Could I grab this one?

HosseinMarzban:
Can I work on this issue?

helfi92:
Fixed in https://github.com/taskcluster/taskcluster-web/commit/2c64f3f8b44c1d7113ded5b59503d4f14d0646c3.

PR #233, opened 2018-10-28
When signing out - clear the Apollo client cache

alecclarke:
RE: https://github.com/taskcluster/taskcluster-web/issues/226

### What does this PR change?
This replaces [`ApolloClient#resetStore`](https://www.apollographql.com/docs/react/api/apollo-client.html#ApolloClient.resetStore) with [`ApolloClient#clearStore`](https://www.apollographql.com/docs/react/api/apollo-client.html#ApolloClient.clearStore) when signing out.

### Why is this important?
When we were calling `ApolloClient#resetStore` - all active queries would be refetched, and if one of those queries returned a response that wasn't parsable as JSON, ApolloClient would throw an error resulting in an incomplete sign out process.

By using `ApolloClient#clearStore`, the store is cleared, but no queries are refetched which allows the sign out process to complete.

### DIFF
--- a/src/components/Dashboard/UserMenu.jsx
+++ b/src/components/Dashboard/UserMenu.jsx
@@ -64,7 +64,7 @@ export default class UserMenu extends Component {
     this.handleMenuClose();
     // Since Apollo caches query results, it’s important to get rid of them
     // when the login state changes.
-    await this.props.client.resetStore();
+    await this.props.client.clearStore();
     this.props.onUnauthorize();
   };
 
--- a/src/components/Dashboard/UserMenu.jsx
+++ b/src/components/Dashboard/UserMenu.jsx
@@ -46,6 +46,9 @@ export default class UserMenu extends Component {
   handleClickSignOut = () => {
     this.handleMenuClose();
     this.props.onUnauthorize();
+    // Since Apollo caches query results, it’s important to get rid of them
+    // when the login state changes.
+    this.props.client.clearStore();
   };
 
   handleMenuClick = e => {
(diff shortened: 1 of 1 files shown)


alecclarke:
@helfi92 Note: the sign-in process seems to complete when on a page ( http://localhost:5080/aws-provisioner/gecko-t-win7-32) with the same error issue, as the `resetStore` call is made after we call [onAuthorize(...)](https://github.com/taskcluster/taskcluster-web/blob/master/src/components/SignInDialog/index.jsx#L70) (Note: in this case https://github.com/taskcluster/taskcluster-web/blob/master/src/components/SignInDialog/index.jsx#L71 doesn't actually look like it's getting hit as the JSON parsable exception is raised first). This is different than the sign out flow as we currently call `resetStore`, then `onUnauthorize()`.

As far as I can see, `clearCache` is still a safe option for signing out as it shouldn't leave any of the previously signed in data cached. Do you still agree?

alecclarke:
Also, I noticed there is a duplicate non-async (`handleClickSignOut`)[https://github.com/taskcluster/taskcluster-web/blob/master/src/components/Dashboard/UserMenu.jsx#L46], is this function in use? (perhaps there is a reason for an async and normal function to be included).

If it's a duplicate, and not needed, I can remove it in a follow up PR.

src/components/Dashboard/UserMenu.jsx
```diff
@@ -64,7 +64,7 @@ export default class UserMenu extends Component {
     this.handleMenuClose();
     // Since Apollo caches query results, it’s important to get rid of them
     // when the login state changes.
-    await this.props.client.resetStore();
+    await this.props.client.clearStore();
     this.props.onUnauthorize();
```
helfi92: Let's remove `async` and have `clearStore` be at the end.

```jsx
handleClickSignOut = () => {
  this.handleMenuClose();
  this.props.onUnauthorize();
  // Since Apollo caches query results, it’s important to get rid of them
  // when the login state changes.
  this.props.client.clearStore();
};
```

PR #244, opened 2018-10-30
Track ConnectionDataTable page number via state

helfi92:
Fixes https://github.com/taskcluster/taskcluster-web/issues/213.

Regarding the issue in https://github.com/taskcluster/taskcluster-web/issues/213., searching for a new client prefix would cause an issue with the page number in `ConnectionDataTable` where page size was not being cleared away.


helfi92:
https://deploy-preview-244--taskcluster-web.netlify.com/auth/clients

djmitche:
Looks good!

Issue #253, opened 2018-11-01
Artifact links all try to display in log viewer


djmitche:
https://taskcluster-web.netlify.com/tasks/A73VbaXFRSCm5LRvKfB4Iw/runs/0/logs/https%3A%2F%2Fqueue.taskcluster.net%2Fv1%2Ftask%2FA73VbaXFRSCm5LRvKfB4Iw%2Fruns%2F0%2Fartifacts%2Fpublic%2Fbuild%2Fsetup.exe

For the most part, I think we should be linking to artifacts in a new browser tab and letting the browser figure out how to display them.  Basically, what tools has been doing all along.

The tools task inspector did a smart thing and privileged `public/logs/live.log` as the "main" log for the run.  It would be great to be able to get to that with one click, without opening artifacts and looking for it (in what may be a long list).

It might be good to do some filename filtering, and if an artifact ends in ".log", perhaps making it link by default to the log viewer.

helfi92:
@djmitche Has your issue been resolved?

djmitche:
I think so!

PR #268, opened 2018-11-05
Open logs in tc-web, otherwise open in new tab

helfi92:
Relates to https://github.com/taskcluster/taskcluster-web/issues/253.

### DIFF
--- a/src/components/TaskRunsCard/index.jsx
+++ b/src/components/TaskRunsCard/index.jsx
@@ -136,7 +136,7 @@ export default class TaskRunsCard extends Component {
     return this.props.runs[this.props.selectedRunId];
   }
 
-  handleArtifactClick = ({ url, isPublicLog }) => {
+  handleArtifactClick = ({ url, isLog }) => {
     if (!url) {
       return null;
     }
@@ -145,7 +145,7 @@ export default class TaskRunsCard extends Component {
       const { history } = this.props;
       const { taskId, runId, state } = this.getCurrentRun();
 
-      if (isPublicLog) {
+      if (isLog) {
         const live = state === 'PENDING' || state === 'RUNNING';
         const encoded = encodeURIComponent(url);
         const path = live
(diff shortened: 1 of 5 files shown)


src/components/TaskRunsCard/index.jsx
```diff
@@ -180,15 +180,27 @@ export default class TaskRunsCard extends Component {
     this.setState({ showArtifacts: !this.state.showArtifacts });
   };
 
+  getLiveLogArtifactFromRun = run => {
+    const artifact = run.artifacts.edges.find(({ node: { name } }) =>
+      name.includes('live_backing.log')
```
helfi92: Is it always `live_backing.log`?

djmitche: No, in fact, that's an implementation detail of docker-worker.

https://docs.taskcluster.net/docs/manual/tasks/artifacts#log-artifacts says:
> By convention, workers record the output of a task -- a command's output to stdout and stderr, for example -- in an artifact named public/logs/live.log. The details of this artifact, such as its storage type and encoding, can differ from worker to worker. You should not assume anything about it, as the implementation may change dynamically to optimize performance.
> 
> While a task is running, some workers can relay data to the client as it is produced, in a single HTTP request. Clients can parse and display this data as it arrives to present a "live log" to the user.

So that's the only well-known filename we should be looking for.

It might be worth amending that documentation to suggest (without making any commitments) that log files should be named with `/logs/` in the filename (per https://github.com/taskcluster/taskcluster-web-server/pull/56), or alternately ending in `.log`, as tools will use that as a hint that the content is a log and should be viewed as such.

helfi92: The UI will now only consider a file to be a log when the contentType starts with `text`. https://github.com/taskcluster/taskcluster-web-server/pull/56/commits/4be56ab750601fe120bda28b1d95cf59202ddad5.

PR #282, opened 2018-11-12
Dynamic favicon based on task group state #270

### LINKED ISSUE #270: Visual indication of a task group state
A user will want to do other things while waiting for a task group to finish. This can be solved by modifying the favicon to have a yellow, green or red circle beside the Logo to indicate the status of a task group. When the user leaves the task group view, we should bring back the original favicon.

projectyang:
Favicon icon will now reflect current status in task group view.

Proposed icons:

Fixes #270


src/views/Tasks/TaskGroup/helmet.jsx
```diff
@@ -0,0 +1,13 @@
+import { Helmet } from 'react-helmet';
+import React, { Component } from 'react';
+
+export default class TestHelmet extends Component {
```
helfi92: I don't think you need to put that in its own file. You can just inline it in the code.

projectyang:
Icon status draft (ignore colour for now). Let me know if sizing/placement is suitable.
untitled 1

helfi92:
Icon looks good to me :)

helfi92:
Actually, I'm not sure anymore. I don't feel like a circle fits well with the logo. What do you think about something like:

screen shot 2018-11-16 at 9 03 05 am

projectyang:
Looks good! Created different sizes incase you were curious. Option 1 looks the best to me.

Option 1
logotest2

screen shot 2018-11-16 at 3 28 31 pm

Option 2
logotest7
screen shot 2018-11-16 at 3 38 43 pm

Option 3
logotest3

screen shot 2018-11-16 at 3 29 20 pm

src/components/Favicon/index.jsx
```diff
@@ -0,0 +1,16 @@
+import { Helmet } from 'react-helmet';
+import React, { Component } from 'react';
+
+export default class Favicon extends Component {
+  render() {
```
helfi92: Let's add the following static methods to `Favicon`:

```jsx
static propTypes = {
  state: taskGroupState,
};

static defaultProps = {
  state: null,
};
```

You can define `taskGroupState` in `src/utils/prop-types` right under `taskState`:

```jsx
export const taskGroupState = oneOf([
  'FAILED',
  'COMPLETED',
  'RUNNING',
]);
```

src/components/Favicon/index.jsx
```diff
@@ -0,0 +1,16 @@
...
+    return (
+      <Helmet>
+        {this.props.status ? (
+          <link href={`/logo${this.props.status}.png`} rel="shortcut icon" />
```
helfi92: Now that status will be uppercased, you can do:

```jsx
import { lowerCase } from 'change-case';

// ...
<link href={`/logo${lowerCase(this.props.state}).png`} rel="shortcut icon" />
```

src/components/TaskGroupProgress/index.jsx
```diff
@@ -244,13 +245,29 @@ export default class TaskGroupProgress extends Component {
     }
   };
 
+  getFaviconStatus = statusCount => {
```
helfi92: Rename `getFaviconStatus` to `getFaviconState`.

helfi92: You can access `statusCount` via `const { statusCount } = this.state;` instead of making it an argument to the method.

src/components/TaskGroupProgress/index.jsx
```diff
@@ -244,13 +245,29 @@ export default class TaskGroupProgress extends Component {
...
+    const { failed, pending, running, unscheduled } = statusCount;
+    const incompletedTasks = pending + running + unscheduled;
+
+    if (failed > 0) {
+      return 'Failed';
```
helfi92: Instead of hard-coding these values, you can use `TASK_STATE.FAILED`, `TASK_STATE.RUNNING`, and `TASK_STATE.COMPLETED`. These can be imported from `src/utils/constants`.

Issue #318, opened 2018-12-18
babel-plugin-transform-react-remove-prop-types


helfi92:
We should consider adding this preset to optimize the production bundle size. Let's add the following plugin for the production environment only:

```js
require.resolve('babel-plugin-transform-react-remove-prop-types'), {
  removeImport: true,
},
```

Issue #54, opened 2018-04-10
Documentation


eliperelman:
In the currently documentation structure, docs are in a separate repo on a separate domain, and its settings are managed independently from taskcluster-web. Let's bring those into this repository, still rendered with Markdown, at the `/docs` endpoint.

PR #319, opened 2018-12-18
[WIP] Raw docs

helfi92:
Resolves https://github.com/taskcluster/taskcluster-web/issues/318.
Refers to https://github.com/taskcluster/taskcluster-web/issues/54.


src/scripts/generateToc.js
```diff
@@ -1,10 +1,25 @@
 const fs = require('fs');
```
djmitche, line 1: Will this script end up being part of `yarn generate`?

Note that `yarn generate` now moves much of the docs data to `generated/` -- that may be a "head start" on getting it imported into the docs site.

djmitche (commented):
This is a *lot* of code, and it's hard to understand the overall approach from this level of detail.  Can you write down the design you're using?

src/scripts/generateToc.js
helfi92: Yes, I believe it should.

> Note that `yarn generate` now moves much of the docs data to `generated/`

That's awesome. Thanks :)

helfi92:
> This is a _lot_ of code, and it's hard to understand the overall approach from this level of detail. Can you write down the design you're using?

Yeah, I figured. I'll write something up soon.

djmitche:
Feel free to change the generation process to put that data somewhere else than generated/ if it's more convenient for you.

PR #322, opened 2018-12-20
Bug 1491070 - run tests and lint in Taskcluster

djmitche:
I tried using `xvfb-run` instead, but it just hung rather than running `yarn test`.

I also included `yarn test` and `yarn lint` in the same task because the long pole is the `yarn` (install) invocation.  Once that's done, the tests and lint are just a few seconds each.


djmitche:
I'll address that error and request review again..

djmitche:
Great success!

.taskcluster.yml
```diff
@@ -0,0 +1,35 @@
+version: 1
+policy:
+  pullRequests: public
+tasks:
+  - taskId: {$eval: as_slugid("tests")}
```
helfi92: Out of curiosity, where is `as_slugid` defined? I wasn't able to find it in https://docs.taskcluster.net/docs/reference/integrations/taskcluster-github/docs/taskcluster-yml-v0.

djmitche: It's in https://docs.taskcluster.net/docs/reference/integrations/taskcluster-github/docs/taskcluster-yml-v1 :)

helfi92: Ahh. Btw, https://tools.taskcluster.net/quickstart defaults to version 0. Should this be updated to version 1?

PR #329, opened 2018-12-22
Show warn popup on hook create close tab

### LINKED ISSUE #320: Warn on exit
**Problem**: https://bugzilla.mozilla.org/show_bug.cgi?id=1515233#c0

**Solution**:

One way we could do t hat is to build a utility component that takes a `message` and a `dirty` flag that alerts the user with the `message` whenever the user attempts to leave the page and the `dirty` flag is set to true. Once this is done, we can start integrating it to the views that need it:

* http://localhost:5080/hooks/create
* http://localhost:5080/secrets/create
* http://localhost:5080/tasks/create
* http://localhost:5080/auth/clients/create
* http://localhost:5080/auth/roles/create
* http://localhost:5080/shell
* http://localhost:5080/display

Adding it to one view is good enough for this issue. We could create a new issue later to have this utility be used by the other views that require this functionality.

arshadkazmi42:
Fixes #320

### DIFF
--- a/src/components/HookForm/index.jsx
+++ b/src/components/HookForm/index.jsx
@@ -32,6 +32,7 @@ import { HOOKS_LAST_FIRE_TYPE } from '../../utils/constants';
 import { hook } from '../../utils/prop-types';
 import removeKeys from '../../utils/removeKeys';
 import ErrorPanel from '../ErrorPanel';
+import WindowCloseWarnPopup from '../../utils/WindowCloseWarnPopup';
 
 const initialHook = {
   metadata: {
@@ -671,6 +672,7 @@ export default class HookForm extends Component {
             }
           />
         )}
+        <WindowCloseWarnPopup />
       </Fragment>
     );
   }
--- a/src/utils/WindowCloseWarnPopup.jsx
+++ b/src/utils/WindowCloseWarnPopup.jsx
@@ -1,0 +1,26 @@
+import { Component } from 'react';
+
+class WindowCloseWarnPopup extends Component {
+  componentDidMount = () => {
+    window.addEventListener('beforeunload', this.keepOnPage);
+  };
+
+  componentWillUnmount = () => {
+    window.removeEventListener('beforeunload', this.keepOnPage);
+  };
+
+  keepOnPage = event => {
+    const message = 'Warning! Unsaved changes';
+
+    // eslint-disable-next-line no-param-reassign
+    event.returnValue = message;
+
+    return message;
+  };
+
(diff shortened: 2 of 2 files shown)


src/utils/WindowCloseWarnPopup.jsx
```diff
@@ -0,0 +1,26 @@
...
+
+  componentWillUnmount = () => {
+    window.removeEventListener('beforeunload', this.keepOnPage);
+  };
+
+  keepOnPage = event => {
+    const message = 'Warning! Unsaved changes';
+
+    // eslint-disable-next-line no-param-reassign
+    event.returnValue = message;
```
helfi92: Custom messages don't seem to be supported that well. In that case let's just return `true` and not worry about the custom message.

src/utils/WindowCloseWarnPopup.jsx
```diff
@@ -0,0 +1,26 @@
...
+
+    return message;
+  };
+
+  render() {
+    return null;
+  }
+}
+
+export default WindowCloseWarnPopup;
```
helfi92: We could probably rewrite as:

```jsx
import React, { Component } from 'react';

// Ask the user for confirmation to leave the page.
// This is useful to use when a page has a form.
export default WrappedComponent =>
  class withAlertOnClose extends Component {
    componentDidMount = () => {
      // A generic string not under the control of the web page
      // will be shown instead of the returned string.
      // Custom text is not well supported yet.
      window.addEventListener('beforeunload', this.keepOnPage);
    };

    componentWillUnmount = () => {
      window.removeEventListener('beforeunload', this.keepOnPage);
    };

    keepOnPage = e => {
      e.preventDefault();
      e.returnValue = true;
    };

    render() {
      return <WrappedComponent {...this.props} />;
    }
  };

```

src/components/HookForm/index.jsx
```diff
@@ -671,6 +672,7 @@ export default class HookForm extends Component {
             }
           />
         )}
+        <WindowCloseWarnPopup />
```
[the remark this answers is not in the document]
helfi92:
```jsx
@withAlertOnClose
@withStyles(...)
export default class  // ...
```

arshadkazmi42:
@helfi92 Thanks for the review.
So as given in the issue, I have only integrated it in hook form create.
We need to add it in other pages also.
So should I create a new PR for those?

helfi92:
You're welcome.

> So should I create a new PR for those?

That would be greatly appreciated :)

PR #396, opened 2019-01-16
[Hooks] Make sure the owner field is an email

ededals:
I modified code to use html5 input type email validation.
As the error message was not specified I added Validation message to helper text.

One thing that is strange though is that also `[a-zA-Z]+@[a-zA-Z]+` gets validated by default.
I am wondering if this is expected behaviour  of html5 input e-mail validation.

I hope I am on the right track :)

### DIFF
--- a/src/components/HookForm/index.jsx
+++ b/src/components/HookForm/index.jsx
@@ -168,6 +168,12 @@ export default class HookForm extends Component {
       scheduleTextField: '',
       taskValidJson: true,
       triggerSchemaValidJson: true,
+      validation: {
+        owner:  {
+          error: false,
+          message: "",
+        },
+      }
     };
   }
 
@@ -344,10 +350,18 @@ export default class HookForm extends Component {
       hook: assocPath(['metadata', 'name'], e.target.value, this.state.hook),
     });
 
-  handleOwnerChange = e =>
+  handleOwnerChange = e =>{
+    debugger;
     this.setState({
       hook: assocPath(['metadata', 'owner'], e.target.value, this.state.hook),
-    });
+      validation:{
+        owner: {
+          error: !e.currentTarget.validity.valid,
+          message: e.currentTarget.validationMessage,
+        },
+      },
+    })
+  };
 
   handleDescriptionChange = e =>
     this.setState({
@@ -375,6 +389,7 @@ export default class HookForm extends Component {
       triggerSchemaInput,
       triggerContextInput,
       hook,
+      validation,
     } = this.state;
     /* eslint-disable-next-line no-underscore-dangle */
     const lastFireTypeName = !isNewHook && hook.status.lastFire.__typename;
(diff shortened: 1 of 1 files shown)


src/components/HookForm/index.jsx
```diff
@@ -168,6 +168,12 @@ export default class HookForm extends Component {
       scheduleTextField: '',
       taskValidJson: true,
       triggerSchemaValidJson: true,
+      validation: {
```
helfi92: Please add `validation: {},` on the initial state object above.

src/components/HookForm/index.jsx
```diff
@@ -168,6 +168,12 @@ export default class HookForm extends Component {
...
+        owner:  {
+          error: false,
+          message: "",
```
helfi92: We use single quotes for strings and double quotes for jsx. Use single quotes here.

src/components/HookForm/index.jsx
```diff
@@ -344,10 +350,17 @@ export default class HookForm extends Component {
       hook: assocPath(['metadata', 'name'], e.target.value, this.state.hook),
     });
 
-  handleOwnerChange = e =>
+  handleOwnerChange = e =>{
```
helfi92: Add space before the braces.

`=> {`.

src/components/HookForm/index.jsx
```diff
@@ -344,10 +350,17 @@ export default class HookForm extends Component {
...
+        owner: {
+          error: !e.currentTarget.validity.valid,
+          message: e.currentTarget.validationMessage,
+        },
+      },
+    })
+  };
 
   handleDescriptionChange = e =>
     this.setState({
```
helfi92: In the `validHook` method, you will want to include `!validation.owner.error`. See the diff.

src/components/HookForm/index.jsx
```diff
@@ -434,9 +448,12 @@ export default class HookForm extends Component {
           </ListItem>
           <ListItem>
             <TextField
+              error = {validation.owner.error}
```
helfi92: Remove spaces before and after the `=` symbol.

src/components/HookForm/index.jsx
ededals: Do I have to add the empty object, or better to instantiate it with owner prop as seen in the comment below. Im asking this because both of the comments are refering to the same code.

helfi92: An empty object is good enough.

src/components/HookForm/index.jsx
```diff
@@ -448,12 +449,16 @@ export default class HookForm extends Component {
...
+                validation.owner !== undefined ? validation.owner.error : false
+              }
               required
               label="Owner Email"
               name="owner"
               type="email"
-              helperText={validation.owner.message}
+              helperText={
+                validation.owner !== undefined ? validation.owner.message : ''
+              }
```
helfi92: `helperText={validation.owner.message}`.

PR #397, opened 2019-01-16
Retain search results on routing back to hooks

### LINKED ISSUE #382: Hooks: going back to the hooks list resets my search
**Summary:**
Clicking away from the hooks list removes the state the page was in, so that when I click my browser's back button, I'm looking at an unfamiliar, fresh version of the page.

**To reproduce:**
1. Go to [hooks](https://taskcluster-web.netlify.com/hooks)
2. Search for `fenix`
3. Click a hook
4. Click your browser's back button

**Preferred behaviour:**
When the browser's back button is pressed, the following state should be remembered:
* Expanded hook groups
* The current search

OjaswinM:
Fixes #382

### DIFF
--- a/src/views/Hooks/ListHooks/index.jsx
+++ b/src/views/Hooks/ListHooks/index.jsx
@@ -6,6 +6,7 @@ import Spinner from '@mozilla-frontend-infra/components/Spinner';
 import { withStyles } from '@material-ui/core/styles';
 import MuiTreeView from 'material-ui-treeview';
 import PlusIcon from 'mdi-react/PlusIcon';
+import * as qs from 'query-string';
 import Dashboard from '../../../components/Dashboard';
 import HelpView from '../../../components/HelpView';
 import Search from '../../../components/Search';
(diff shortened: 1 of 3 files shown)


OjaswinM:
Hey, i made the requested changes! Also, any specific reason of using qs over query-string?

helfi92:
The reason is primarily for its support for nesting query strings which we might use in the future on one of our views.

helfi92:
> When I tried using <search value={hookSearch}> I was unable to change it in the gui so I used placeholder instead. Is it okay?

Not really. We should make sure it works with `value`. Try looking into the Search component to see what's not working.

helfi92:
That's a very smart idea. I'll give it a try in a bit.

src/views/Hooks/ListHooks/index.jsx
```diff
@@ -64,6 +77,7 @@ export default class ListHooks extends Component {
         search={
           <Search
             placeholder="Hook contains"
+            defaultValue={hookSearch || null}
```
helfi92: `defaultValue={hookSearch}`

src/views/Hooks/ListHooks/index.jsx
```diff
@@ -29,12 +30,24 @@ export default class ListHooks extends Component {
...
+  }
+
   handleCreateHook = () => {
     this.props.history.push('/hooks/create');
   };
 
   handleHookSearchSubmit = hookSearch => {
-    this.setState({ hookSearch });
+    this.props.history.push(`/hooks?search=${hookSearch}`);
   };
```
helfi92:
```jsx
handleHookSearchSubmit = hookSearch =>
  this.props.history.push(
    `/hooks${hookSearch ? `?search=${hookSearch}` : ''}`
  );
```

PR #413, opened 2019-01-24
Bug 1520857- add support in tc-web for displaying information from the lastFire table

Biboswan:


Biboswan:
hook1
hook3

The error panel slighty goes out in very small screens as seen in the 2nd pic. When fixing the width there were some other very bad overlapping with other fired hooks info
cc @djmitche

src/views/Hooks/ViewHook/index.jsx
```diff
@@ -10,18 +10,32 @@ import createHookQuery from './createHook.graphql';
...
+  graphql(hookLastFiresQuery, {
+    skip: ({ match: { params } }) => !params.hookId,
+    options: ({ match: { params } }) => ({
+      variables: {
+        hookGroupId: params.hookGroupId,
+        hookId: decodeURIComponent(params.hookId),
+      },
+    }),
+    name: 'lastFiresData',
   }),
```
helfi92: No need to create `hookLastFires.graphql`. You should be able to add everything in `hook.graphql`.

src/views/Hooks/ViewHook/index.jsx
```diff
@@ -128,13 +142,13 @@ export default class ViewHook extends Component {
   };
 
   render() {
-    const { isNewHook, data } = this.props;
+    const { isNewHook, hookData, lastFiresData } = this.props;
```
helfi92: One you merge the last fires data in `hook.graphql` then everything will be under `data`.

src/views/Hooks/ViewHook/hookLastFires.graphql
```diff
@@ -0,0 +1,9 @@
+query HookLastFires($hookGroupId: ID!, $hookId: ID!) {
```
helfi92: Delete this file. See reason in other comment.

src/components/HookForm/index.jsx
```diff
@@ -508,61 +531,85 @@ export default class HookForm extends Component {
...
+                    <Typography variant="subtitle1">
+                      Last Hook Fire Attempts
+                    </Typography>
                   }
                 />
-              </ListItem>
-            </Fragment>
+                {hookLastFires.map(
+                  ({ taskId, taskCreateTime, firedBy, result, error }) => (
+                    <ListItem key={taskId}>
```
helfi92: Let's use the `DataTable` component in `src/components/DataTable` to display the result? The table will be  similar to something like http://localhost:5080/provisioners/aws-provisioner-v1.

src/components/HookForm/index.jsx
```diff
@@ -131,6 +139,8 @@ export default class HookForm extends Component {
   static propTypes = {
     /** A GraphQL hook response. Not needed when creating a new hook  */
     hook: hook.isRequired,
+    /** Object */
```
helfi92: Let's use a different comment here to describe the property.

src/components/HookForm/index.jsx
```diff
@@ -508,61 +531,85 @@ export default class HookForm extends Component {
...
+                                </Grid>
+                                <Grid item>
+                                  <Label mini status={result.toLowerCase()}>
+                                    {result}
+                                  </Label>
+                                </Grid>
+                              </Grid>
+                            </ExpansionPanelSummary>
+                            <ExpansionPanelDetails>
+                              <ErrorPanel error={error} onClose={null} />
```
helfi92: We could probably have errors be a separate column in the table and use simple text to display the error instead of having it in a error panel. We use error panels for when the server fails to return data or the user is trying to do something not allowed.

helfi92 (changes requested):
Good progress :) Added some requested changes.

src/components/HookForm/index.jsx
Biboswan: I don't think a DataTable would be a good fit for the data. If result is  `success` then expandable panel has nothing to show. If result is `error` then there is very large error text  to show on opening the panel. In  any column it could be trouble to fit the error text especially.

Biboswan:
lastfires

Biboswan:
have a look @helfi92

helfi92:
Awesome, I'll have a look soon.

owlishDeveloper:
@helfi92 - I'm a bit out of the loop on the plans for this repository. It's not being archived after having merged to the monorepo?

helfi92:
@owlishDeveloper I unarchived the repo so @Biboswan and I can continue iterating on this PR. Once we are happy with the PR, @Biboswan will create a PR on on the monorepo and we will re-archive this repo. The discussion happened in [Bug 1520857](https://bugzilla.mozilla.org/show_bug.cgi?id=1520857#c10).

src/components/DataTable/index.jsx
```diff
@@ -7,13 +7,15 @@ import {
   oneOf,
   oneOfType,
   object,
+  bool,
```
helfi92, line 10: Why are we adding pagination? Is the last fire endpoint paginated (i.e., uses continuation token)?

src/components/HookForm/index.jsx
```diff
@@ -508,61 +538,76 @@ export default class HookForm extends Component {
...
+                items={hookLastFires}
+                headers={['Task ID', 'FiredBy', 'Result', 'Attempted', 'Error']}
+                isPaginate
+                renderRow={hookFire => (
+                  <TableRow key={hookFire.taskId}>
+                    <TableCell>
+                      {(hookFire.result === 'SUCCESS' && (
+                        <Link to={`/tasks/${hookFire.taskId}`}>
+                          <Typography>{hookFire.taskId}</Typography>
+                        </Link>
```
helfi92: Right now it looks like:

screen shot 2019-02-07 at 1 22 09 pm

It should look like this:

screen shot 2019-02-07 at 1 23 11 pm

This will make it look like any other link in a table to the rest of the app. Here's an example: [The Worker ID column in the Workers view](http://localhost:5080/provisioners/aws-provisioner-v1/worker-types/android-components-g).

src/components/HookForm/index.jsx
```diff
@@ -508,61 +538,76 @@ export default class HookForm extends Component {
...
+                          </ExpansionPanelSummary>
+                          <ExpansionPanelDetails>
+                            <ErrorPanel
+                              className={classes.errorPanel}
+                              error={hookFire.error}
+                              onClose={null}
+                            />
+                          </ExpansionPanelDetails>
+                        </ExpansionPanel>
+                      )) || <Typography>N/A</Typography>}
```
helfi92: Use lower case `n/a`.

src/components/HookForm/index.jsx
```diff
@@ -508,61 +538,76 @@ export default class HookForm extends Component {
...
+                      <Label mini status={hookFire.result.toLowerCase()}>
+                        {hookFire.result}
+                      </Label>
+                    </TableCell>
+                    <TableCell>
+                      <DateDistance from={hookFire.taskCreateTime} />
+                    </TableCell>
+                    <TableCell className={classes.errorTableCell}>
+                      {(hookFire.result === 'ERROR' && (
+                        <ExpansionPanel>
```
helfi92: Is there a hook I can see where `hookFire.result === 'ERROR'`? I want to see how it looks in the UI.

src/views/Hooks/ViewHook/index.jsx
```diff
@@ -152,6 +152,10 @@ export default class ViewHook extends Component {
                 dialogError={dialogError}
                 actionLoading={actionLoading}
                 hook={data.hook}
+                hookLastFires={data.hookLastFires.sort(
+                  (a, b) =>
+                    new Date(b.taskCreateTime) - new Date(a.taskCreateTime)
+                )}
```
helfi92: Extract logic from jsx:

```jsx
const hookLastFires = // ...

// ...
hookLastFires={hookLastFires}
```

helfi92 (changes requested):

src/components/DataTable/index.jsx
Biboswan: As per discussion with Dustin in irc we are keeping as many as 100 lastfire info rows for each hookId (expiration logic). Hence I think showing all rows at once won't be good.

src/components/HookForm/index.jsx
Biboswan: If the lastfire result is `success` then the taskId links to another route. But if result is `error` then no task of the mentioned `taskId` is actually created hence there is no link to the corresponding task. So I'm underlinking taskId for the sucess ones and no underline for `error` ones to distinguish them. Shall I change it ?

Biboswan:
lastfire3

src/components/DataTable/index.jsx
djmitche: We also only purge those once per day, so a particularly active hook might have many more rows than that at some time during the day!

helfi92:
@Biboswan How can I get to see the list of errors in the UI when I pull the code down locally? I want to try to make some tweaks on my local machine.

Biboswan:
@helfi92  generally hooks with `hookGroup` "project-mobile" have a lot of error and success. Sorry I missed your previous comment on that

helfi92:
Thanks! I'll take a look.

src/components/HookForm/index.jsx
helfi92: Instead of having an expansion panel for each error, we should instead have a button that opens a drawer with the error on click. This is similar to https://taskcluster-web.netlify.com/provisioners/aws-provisioner-v1 when clicking the info icon. The drawer can display the `ErrorPanel` component.

screen shot 2019-02-11 at 10 59 46 am

src/components/HookForm/index.jsx
helfi92: From https://github.com/taskcluster/taskcluster-web/pull/413/files#r254810845, we could remove the link icon when the task has an error.

helfi92 (changes requested):
Added requested changes.

Biboswan:
t
 Only error message i could find at the moment

helfi92 (changes requested):
Actual: The error is shown in the table
Expected: The error is displayed  outside the table in a `Drawer` component. To see what I mean:

1. Navigate to https://taskcluster-web.netlify.com/provisioners/aws-provisioner-v1
2. Click on the info icon. You will see a [drawer](https://material-ui.com/demos/drawers/) that will slide from the right hand side of the screen.

_the info button in the worker types view_
screen shot 2019-02-19 at 3 04 57 pm

_the drawer in the worker type view after clicking the info button_

screen shot 2019-02-19 at 3 08 44 pm


Coming back to this issue, we want to display the error inside the drawer, not inside the table. You can look at the code of https://taskcluster-web.netlify.com/provisioners/aws-provisioner-v1 to see how the drawer is implemented.

Hope that helps. Let me know if you had any questions/concerns.

Biboswan:
ah i just skipped the drawer :(. making the changes

Biboswan:
screenshot from 2019-02-22 23-19-04

helfi92 (changes requested):
Looks a lot better! A couple of points:

* https://github.com/taskcluster/taskcluster-web/pull/413/files#r254810845 is not done yet.
* Let's show "n/a" (without the quotes) instead of the information icon when there's no error.

Biboswan:
> * https://github.com/taskcluster/taskcluster-web/pull/413/files#r254810845 is not done yet.

Did I misunderstood something. When the task has an error there is no link icon aka underline as you said.When there is no error the task exists so taskId has the link icon as `/tasks/${hookFire.taskId}` exists. So do you want me to remove link in both the cases ?

helfi92:
> When the task has an error there is no link icon aka underline as you said.When there is no error the task exists so taskId has the link icon as `/tasks/${hookFire.taskId}` exists. So do you want me to remove link in both the cases ?

Instead of the underline, it should match the following style (look at the link icon beside the id): https://user-images.githubusercontent.com/3766511/52433650-8ccd9b00-2adb-11e9-9df7-a5e272407a82.png. There is also a hover effect that comes from `theme.mixins.listItemButton`.

helfi92 (changes requested):
Nice improvements :) A couple of points:

* A task ID that has no link should not be wrapped in a TableCellListItem since it shouldn't be a list item. It should be just regular text.

screen shot 2019-02-25 at 9 38 30 am

* There seems to be `))` beside the information icon


screen shot 2019-02-25 at 9 38 02 am

helfi92 (changes requested):
A couple of changes:

1. The "<" and ">" buttons should be to the right of the table, similar to how they are show in https://material-ui.com/demos/tables/#tables.

screen shot 2019-02-25 at 3 03 50 pm

2. Set `rowsPerPage` to 5

3. Remove the option to change row count

screen shot 2019-02-25 at 3 09 00 pm

4. Rename "Last Hook Fire Attempts" to "Last Fired Results"

5. The `FiredBy` column should be shown as `Fired By`. You can use the `change-case` package to modify the title. See how we do it for the other tables in the app.

Biboswan:
@helfi92 you want to shift just the '< >' buttons rightward and not the page no. (1-20) part ?

helfi92:
If we can remove the (1-20) part then that would be the best because it will make it consistent with the other tables we use on the site.

Biboswan:
i was asking whether to shift right both arrows and 1-20 or only arrows. If we remove 1-20 then user will not know total count is that ok

helfi92:
> i was asking whether to shift right both arrows and 1-20 or only arrows. If we remove 1-20 then user will not know total count is that ok

Yes that's okay. We want to maintain consistency across our tables. None of our tables that have that (1-20) extra information.

helfi92:
How is it going @Biboswan. I am back now :)

Biboswan:
Few days ago when I tried to resume the work I got errors on running the  web server. I tried both `Yarn start` and `Yarn start devSever`

helfi92:
I pushed https://github.com/taskcluster/taskcluster/commit/849447681e4ab2e44934ebf94494f5d1b4dde7af today. Try pulling the recent changes and see if that fixes it :)

Biboswan:
it's done will push today

Biboswan:
t

src/components/DataTable/index.jsx
```diff
@@ -7,14 +7,27 @@ import {
...
 import TableRow from '@material-ui/core/TableRow';
+import TablePagination from '@material-ui/core/TablePagination';
 
+@withStyles(() => ({
+  noDisplay: {
+    display: 'none',
+  },
+  tablePaginationRoot: {
+    position: 'relative',
+    left: '207%',
```
helfi92: I don't think we need this. You can look at how they do it in https://material-ui.com/demos/tables/.

src/components/DataTable/index.jsx
```diff
@@ -115,9 +145,36 @@ export default class DataTable extends Component {
...
-            items.map(renderRow)
+            items
+              .slice(page * rowsPerPage, page * rowsPerPage + rowsPerPage)
+              .map(renderRow)
           )}
         </TableBody>
+        {isPaginate && (
+          <TableFooter>
+            <TableRow>
+              <TablePagination
```
helfi92: Screen Shot 2019-03-13 at 9 03 49 AM

We should:
* Have the TablePagination outside the table. This would make sure the pagination is always displayed to the user even when the table is being viewed on a small screen where scrolling is required to view the data. There's an example of that in https://material-ui.com/demos/tables/#sorting-amp-selecting. Basically you will need to remove `TableFooter` and `TableRow` and then have `TablePagination` outside the table.
* The table should be scrollable similar to how all the tables are in the material site.

src/components/HookForm/index.jsx
```diff
@@ -114,12 +123,37 @@ const initialHook = {
...
     fontSize: theme.typography.pxToRem(16),
   },
+  displayBlock: {
+    display: 'block',
+  },
+  errorTableCell: {
+    whiteSpace: 'normal',
+  },
+  errorPanel: {
+    maxHeight: '300px',
```
helfi92: No need to put quotations when the units are `px`.

`maxHeight: 300,`

helfi92 (changes requested):
Good progress!

Biboswan:
d

Biboswan:
d3
Which one is more appealing?

helfi92:
Second one.

helfi92 (commented):
@Biboswan I don't seem to be able to open a hook for some reason. Anyways, I think we are in a better shape now to open a PR under [taskcluster/taskcluster](http://github.com/taskcluster/taskcluster). Do you want to do that?

Biboswan:
Yes! Will do that.

